PR #34, opened 2025-12-05
Fix SessionStart hook with robust wrapper script

EthanJStark:
## Summary

Fixes SessionStart hook that was failing with "operation was aborted" error.

## Changes

- Add `hooks/run-hook.cmd` - Cross-platform polyglot wrapper (borrowed from superpowers plugin)
- Add `hooks/session-start-sync.sh` - Bash wrapper that dynamically resolves plugin root
- Update `hooks/hooks.json` to use `${CLAUDE_PLUGIN_ROOT}/hooks/run-hook.cmd session-start-sync.sh`
- Remove `async: true` flag (not needed)
- Remove `--background` CLI flag (doesn't actually background)

## Problem

The original hook used relative path `../cli/episodic-memory sync --background` which:
1. Assumed hook executes from `hooks/` directory (incorrect)
2. Used `--background` flag that prints message but doesn't actually fork

Result: Hook appeared to succeed but never synced conversations.

## Solution

Use the proven pattern from superpowers plugin:
1. Claude Code sets `CLAUDE_PLUGIN_ROOT` environment variable when executing plugin hooks
2. `run-hook.cmd` is a polyglot wrapper that works on Windows/macOS/Linux
3. Wrapper script resolves plugin root dynamically using bash directory resolution
4. Sync runs synchronously (~5-20 seconds, acceptable for session start)

**Important:** The command path should NOT be quoted. While some plugins use quotes around `${CLAUDE_PLUGIN_ROOT}`, this plugin's hook system passes quotes literally, causing execution to fail.

## Testing

Verified with actual Claude Code session restart:
- ✅ Hook executes at session start without "operation aborted" error
- ✅ Database file modified timestamp matches session restart time
- ✅ Exchange count increases after restart (2,903 → 2,905 confirmed)
- ✅ Sync completes successfully  
- ✅ Works regardless of working directory
- ✅ No session start delays

## Root Cause Analysis

Discovered via systematic debugging:
- `${CLAUDE_PLUGIN_ROOT}` variable expansion works in hook commands ✅
- Quotes around path are passed literally and break execution ❌
- Unquoted path works correctly ✅
- CLI `--background` flag doesn't actually fork background process ❌
- Synchronous sync completes fast enough for session start ✅

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -6,8 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "${CLAUDE_PLUGIN_ROOT}/cli/episodic-memory sync --background",
-            "async": true
+            "command": "${CLAUDE_PLUGIN_ROOT}/hooks/run-hook.cmd session-start-sync.sh"
           }
         ]
       }
--- a/hooks/run-hook.cmd
+++ b/hooks/run-hook.cmd
@@ -1,0 +1,19 @@
+: << 'CMDBLOCK'
+@echo off
+REM Polyglot wrapper: runs .sh scripts cross-platform
+REM Usage: run-hook.cmd <script-name> [args...]
+REM The script should be in the same directory as this wrapper
+
+if "%~1"=="" (
+    echo run-hook.cmd: missing script name >&2
+    exit /b 1
+)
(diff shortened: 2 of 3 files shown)


EthanJStark:
Built this PR, then found and tested #25 , confirmed it's working and a better approach than mine.

PR #63, opened 2026-02-24
feat: add metadata filters (project, session_id, git_branch) to search

jwk2601:
## Summary

The `exchanges` table already has indexed columns for `project`, `session_id`, and `git_branch`, but the search API only exposes time-based filters (`after`/`before`). This PR adds metadata filtering to both the MCP tool and the CLI, enabling project-specific and branch-specific conversation searches.

**Motivation:** Multi-project users frequently need to scope search results to a specific project or git branch. The data and indexes already exist in the database — this change simply exposes them through the existing search interfaces.

### `src/search.ts`
- Extended `SearchOptions` interface with `project`, `session_id`, `git_branch`
- Added `validateMetadataFilter()` — regex validation + length check, matching the existing `validateISODate()` pattern
- Unified `timeFilter` array → general `filters` array for WHERE clause construction
- Added 3x over-fetch for vector search when metadata filters are active (vec0 virtual table applies KNN before WHERE post-filter)
- Added `session_id`, `git_branch` to SELECT columns and result mapping

### `src/mcp-server.ts`
- Added Zod schema entries for new parameters (with `min(1)` + `.optional()`)
- Added JSON schema entries in `ListToolsRequestSchema` handler
- Passed new parameters through in both single and multi-concept search paths

### `src/search-cli.ts`
- Added `--project`, `--session-id`, `--git-branch` CLI flags
- Updated help text with new options and examples

### `test/integration.test.ts`
- Added `describe('Metadata Filtering')` block with 9 tests:
  - Filter by project / non-matching project returns empty / filter by git_branch / filter by session_id
  - Combined metadata + time filters
  - SQL injection rejection / overly long input rejection
  - Vector search mode with metadata filter
  - Backward compatibility (no filter = same results)

## Design decisions

- **Input validation:** Follows the existing pattern of regex validation + string interpolation (same as `validateISODate`). A full parameterized query refactor is out of scope for this PR.
- **Over-fetch factor (3x):** Conservative multiplier to compensate for vec0's KNN-first behavior when metadata filters reduce the result set. The results are trimmed back to the requested limit after filtering.
- **Exact match only:** All three filters use `=` (exact match). Prefix/fuzzy matching can be added in a follow-up if needed.

## Backward compatibility

All new parameters are optional. When omitted, behavior is identical to the current version — the `filters` array remains empty and produces the same SQL as the previous `timeFilter` approach.

## Test plan

- [x] `npm run build` — compiles without errors
- [x] `npm test` — all new tests pass; no regressions in existing tests
- [x] Existing tests with no filters behave identically

## Release Notes

* **New Features**
  * Added metadata filtering to search: users can now filter results by project name, session ID, and git branch.
  * Search results now include session ID and git branch information for better context and traceability.
  * CLI now supports --project, --session-id, and --git-branch filter options for refined queries.
  * Input validation ensures filter values meet safety requirements.

### DIFF
--- a/dist/mcp-server.js
+++ b/dist/mcp-server.js
@@ -4725,6 +4725,7 @@ var require_pattern = __commonJS({
     "use strict";
     Object.defineProperty(exports, "__esModule", { value: true });
     var code_1 = require_code2();
+    var util_1 = require_util();
     var codegen_1 = require_codegen();
     var error2 = {
       message: ({ schemaCode }) => (0, codegen_1.str)`must match pattern "${schemaCode}"`,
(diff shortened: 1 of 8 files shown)


obra:
Hi @jwk2601 — landed an equivalent feature in f061b50, just auto-closed via the commit message.

A few things changed in the re-derivation:

- **Bound parameters instead of regex-validated string interpolation.** Your PR added a `SAFE_STRING_REGEX` and interpolated values into the WHERE fragment, which works (the regex disallows quote chars) but is one regex relaxation away from a SQL injection bug. The new version binds every filter value as a positional parameter, including the existing `after`/`before` clauses that previously interpolated. SQL injection is unreachable by construction; no input scrubbing is required.
- **Same vec0 over-fetch heuristic.** Kept the `limit * 3` candidate window when metadata filters are active. That insight was yours.
- **`searchMultipleConcepts` works automatically.** It already spreads its options through to `searchConversations`, so the new fields flow through without separate plumbing.
- **TDD-derived test set.** Slightly different from your 9 tests — covers per-field filters, combined AND semantics, empty-result, an explicit single-quote injection attempt, and an unfiltered regression.

Your PR's design and your honest scoping comment ("a full parameterized query refactor is out of scope for this PR") were both correct. The refactor turned out to be small enough to do in this round, so I rolled it in.

PR #73, opened 2026-03-28
fix: Delete tool_calls before exchanges to prevent FK constraint failure

teehemkay:
## Summary

- `deleteExchange()` deleted from `vec_exchanges` and `exchanges` but not from `tool_calls`, which has a `FOREIGN KEY (exchange_id) REFERENCES exchanges(id)` constraint
- `index --repair` crashed with `SQLITE_CONSTRAINT_FOREIGNKEY` when removing orphaned entries that had associated tool calls
- Added `DELETE FROM tool_calls WHERE exchange_id = ?` before the exchange deletion

## Test plan

- [x] Run `episodic-memory index --repair` on a database with orphaned entries that have tool calls — previously crashed, now completes successfully

### DIFF
--- a/dist/db.js
+++ b/dist/db.js
@@ -158,6 +158,8 @@ export function getFileLastIndexed(db, archivePath) {
 export function deleteExchange(db, id) {
     // Delete from vector table
     db.prepare(`DELETE FROM vec_exchanges WHERE id = ?`).run(id);
+    // Delete from tool_calls table (must come before exchanges due to FK constraint)
+    db.prepare(`DELETE FROM tool_calls WHERE exchange_id = ?`).run(id);
     // Delete from main table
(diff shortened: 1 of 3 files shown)


obra:
6000 lines of diff is a lot for a four line summary that claims to fix a constraint failure. What's up?

obra:
Hi @teehemkay — thanks for diagnosing this. Closing in favor of 61d7eb0, which takes the architectural route: tool_calls now declares `ON DELETE CASCADE`, and existing databases get a one-time migration that recreates the table with cascade and drops any pre-existing orphan rows. With the schema enforcing the invariant, deleteExchange() (and any future deletion path) is safe by construction.

Your one-line manual fix would have worked for the immediate crash; this version makes the bug class unreachable and cleans up the orphans that already accumulated. The diagnosis was yours — credit for finding it.

Closes #81.

PR #77, opened 2026-04-09
fix: convert sqlite-vec distances into cosine similarity scores

### LINKED ISSUE #55: Similarity score calculation bug in src/search.ts - L2 distance treated as cosine distance
The similarity score calculation in src/search.ts line 139 produces incorrect results (negative percentages, meaningless rankings) because it treats the Euclidean (L2) distance returned by sqlite-vec as if it were cosine distance.

**Current code (src/search.ts, line 139):**
```typescript
similarity: mode === 'text' ? undefined : 1 - row.distance,
```

**Problem:**
sqlite-vec returns L2 (Euclidean) distance, not cosine distance. For normalized embedding vectors, the relationship between L2 distance `d` and cosine similarity `s` is:

```
s = 1 - (d^2 / 2)
```

The current formula `1 - d` produces values that go negative for any distance > 1, which is common with L2 distances on 384-dimensional vectors. In practice, this means almost every search result shows a negative or near-zero similarity score.

**Fix:**
```typescript
similarity: mode === 'text' ? undefined : 1 - (row.distance * row.distance / 2),
```

**Before fix:** Scores like 0%, -5%, -11%, -16%
**After fix:** Scores like 50%, 45%, 39%, 32%

The ranking order of results was unaffected (L2 and cosine ordering are monotonically related for normalized vectors), but the scores were meaningless and confusing to users.

Tested on v1.0.15 with all-MiniLM-L6-v2 embeddings against ~235 indexed conversations.

officialasishkumar:
## Summary
- convert normalized sqlite-vec L2 distances back into cosine similarity scores before formatting search results
- clamp the converted score into the valid `[-1, 1]` range to absorb tiny floating-point spillover
- add focused coverage for the distance-to-similarity helper and update the integration expectation to match the corrected score range

## Validation
- `npx vitest run test/search.test.ts`
- `npm run build`

Closes #55

### DIFF
--- a/dist/search.d.ts
+++ b/dist/search.d.ts
@@ -5,6 +5,7 @@ export interface SearchOptions {
     after?: string;
     before?: string;
 }
+export declare function normalizedL2DistanceToSimilarity(distance: number): number;
 export declare function searchConversations(query: string, options?: SearchOptions): Promise<SearchResult[]>;
 export declare function formatResults(results: Array<SearchResult & {
     summary?: string;
--- a/dist/search.js
+++ b/dist/search.js
@@ -2,6 +2,10 @@ import { initDatabase } from './db.js';
 import { initEmbeddings, generateEmbedding } from './embeddings.js';
 import fs from 'fs';
 import readline from 'readline';
+export function normalizedL2DistanceToSimilarity(distance) {
+    const similarity = 1 - ((distance * distance) / 2);
+    return Math.max(-1, Math.min(1, similarity));
+}
 function validateISODate(dateStr, paramName) {
     const isoDateRegex = /^\d{4}-\d{2}-\d{2}$/;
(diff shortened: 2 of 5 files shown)


obra:
Hi @officialasishkumar — thanks for diagnosing this and proposing the fix. We landed an equivalent change in 0b5f02f, re-derived via TDD: unit tests for the L2→cosine math (test/cosine-similarity.test.ts) before wiring into searchConversations, and the existing multi-concept test got a relative-ranking assertion instead of the absolute < 0.1 threshold (which worked only because the old broken formula produced negatives for orthogonal-ish text — calibrated against the bug, not against cosine semantics).

Your diagnosis was correct and your formula was correct. Closing this PR in favor of the merged version, but the credit is yours for finding it. Same for #78 — appreciate both contributions.

PR #78, opened 2026-04-09
fix: quote the SessionStart hook path for Windows home directories with spaces

### LINKED ISSUE #75: SessionStart hook breaks on Windows when CLAUDE_PLUGIN_ROOT contains spaces
## Bug

The `SessionStart` hook in `hooks/hooks.json` fails on Windows when the user's home directory contains spaces (e.g., `C:\Users\First Last\...`).

## Root Cause

The hook command doesn't quote `${CLAUDE_PLUGIN_ROOT}`:

```json
"command": "node ${CLAUDE_PLUGIN_ROOT}/cli/episodic-memory.js sync --background"
```

When `CLAUDE_PLUGIN_ROOT` expands to a path with spaces, the shell word-splits it:

```
$ bash -c "node ${CLAUDE_PLUGIN_ROOT}/cli/episodic-memory.js sync --background"
# node receives "/c/Users/First" as the script path, which doesn't exist
```

## Impact

- SessionStart hook error on every new Claude Code session
- Affects any Windows user whose username contains a space (very common default)

## Fix

Quote the variable in `hooks/hooks.json`:

```json
"command": "node \"${CLAUDE_PLUGIN_ROOT}/cli/episodic-memory.js\" sync --background"
```

For reference, the `superpowers` plugin handles this correctly:
```json
"command": "\"${CLAUDE_PLUGIN_ROOT}/hooks/run-hook.cmd\" session-start"
```

## Environment

- Windows 11 Pro
- Claude Code CLI
- episodic-memory plugin v1.0.15
- User home directory: `C:\Users\First Last` (space in path)

officialasishkumar:
## Summary
- quote `${CLAUDE_PLUGIN_ROOT}/cli/episodic-memory.js` in `hooks/hooks.json` so the SessionStart hook survives shell word-splitting on Windows paths with spaces
- add a regression test that loads the hook config and asserts the command stays quoted

## Validation
- `npx vitest run test/hooks.test.ts`
- `jq . hooks/hooks.json >/dev/null`

Closes #75

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node ${CLAUDE_PLUGIN_ROOT}/cli/episodic-memory.js sync --background",
+            "command": "node \"${CLAUDE_PLUGIN_ROOT}/cli/episodic-memory.js\" sync --background",
             "async": true
           }
         ]
--- a/test/hooks.test.ts
+++ b/test/hooks.test.ts
@@ -1,0 +1,14 @@
+import { describe, expect, it } from 'vitest';
+import { readFileSync } from 'fs';
+
+describe('plugin hook configuration', () => {
+  it('quotes CLAUDE_PLUGIN_ROOT in the SessionStart command', () => {
+    const hooks = JSON.parse(
+      readFileSync(new URL('../hooks/hooks.json', import.meta.url), 'utf-8')
+    );
+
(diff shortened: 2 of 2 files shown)


obra:
Thanks @officialasishkumar! Merged. The quote fix is the right minimal change — the bug isn't strictly Windows-specific (any plugin path with spaces would hit it), and the regression test is a useful canary against silently regressing the JSON value.

PR #85, opened 2026-04-28
fix: use high-water mark for incremental indexing (fixes #84)

### LINKED ISSUE #84: Indexer skips entire file when any rows exist, leaving appended turns permanently unindexed
## Summary

`src/indexer.ts:294-298` checks whether *any* rows already exist for a given `archive_path` and skips the entire file if so. There is no comparison against mtime or `MAX(line_end)`. Any `.jsonl` transcript that gets appended to after its first indexing pass — which is the normal case for resumed sessions, or any long session where a sibling Claude Code session triggered a `SessionStart` sync mid-way — has its tail permanently excluded from the index.

Sync still copies the updated archive correctly (mtime check works); only the indexer is at fault.

## Evidence (real index, plugin v1.0.15)

| Metric | Value |
|---|---|
| Total archived conversations | 2,361 |
| Files with fresh index | 1,173 (49.7%) |
| Files with **unindexed tail content** | **1,188 (50.3%)** |
| Total lines on disk | 394,291 |
| Total lines indexed | 386,526 |
| Total lines **unindexed** | **7,765 (2.0%)** |

Distribution of per-file unindexed-line delta (stale files only):

| Stat | Lines |
|---|---|
| median | 5 |
| p95 | 12 |
| max | 1,308 |
| files with delta ≥ 50 | 5 |
| files with delta ≥ 100 | 3 |
| files with delta ≥ 1000 | 1 |

So the typical case is "last few turns of a session never made it in" (consistent with a `SessionStart` background sync racing the still-running session that produced it). The long tail is the painful case — large multi-hour sessions where most of the conversation is silently missing from semantic search.

## Detection query

For anyone wanting to check their own index:

```python
import sqlite3, os
con = sqlite3.connect(os.path.expanduser("~/.config/superpowers/conversation-index/db.sqlite"))
rows = con.execute("SELECT archive_path, MAX(line_end) FROM exchanges GROUP BY archive_path").fetchall()
for path, last in rows:
    if not os.path.exists(path): continue
    with open(path, 'rb') as f: n = sum(1 for _ in f)
    if n > (last or 0):
        print(n - last, path)
```

## Root cause (current behavior)

```ts
// src/indexer.ts:294
const alreadyIndexed = db.prepare(
  'SELECT COUNT(*) as count FROM exchanges WHERE archive_path = ?'
).get(archivePath);
if (alreadyIndexed.count > 0) continue;
```

## Proposed fix

The schema already stores `line_start` and `line_end` per exchange, so the data model supports incremental indexing without a migration. Replace the boolean skip with a high-water-mark check:

1. `SELECT COALESCE(MAX(line_end), 0) FROM exchanges WHERE archive_path = ?`
2. Parse the file and only index exchanges whose `line_start > maxIndexedLine`
3. Embed/insert only the tail; existing rows untouched

The unused `last_indexed` column hints this was the original intent.

## Workaround until fixed

Delete affected rows so the next sync re-indexes from scratch (re-embeds the whole file — non-trivial cost on a large archive):

```sql
DELETE FROM exchanges WHERE archive_path IN (...stale paths...);
```

Happy to open a PR if you'd like — the change is small and well-scoped.

anastasiiaanfimova:
## Problem

The indexer skips a file entirely if `COUNT(*) > 0` for its `archive_path`:

```ts
const alreadyIndexed = db.prepare(
  'SELECT COUNT(*) as count FROM exchanges WHERE archive_path = ?'
).get(archivePath) as { count: number };

if (alreadyIndexed.count > 0) continue;
```

This means any conversation that was indexed in a previous run will never receive new exchanges — even if the source `.jsonl` file has grown since then. Fixes #84.

## Fix

Replace the count check with a high-water mark using `MAX(line_end)`. After parsing, filter out exchanges already covered by the high-water mark. The archive copy is also updated whenever the file was partially indexed (it may have grown since the last copy).

```ts
const maxIndexedResult = db.prepare(
  'SELECT COALESCE(MAX(line_end), 0) as max_line FROM exchanges WHERE archive_path = ?'
).get(archivePath) as { max_line: number };
const maxIndexedLine = maxIndexedResult.max_line;

if (!fs.existsSync(archivePath) || maxIndexedLine > 0) {
  fs.copyFileSync(sourcePath, archivePath);
}

const exchanges = await parseConversation(sourcePath, project, archivePath);
if (exchanges.length === 0) continue;

const newExchanges = maxIndexedLine > 0
  ? exchanges.filter(e => e.lineStart > maxIndexedLine)
  : exchanges;

if (newExchanges.length === 0) continue;

unprocessed.push({ ...conv, exchanges: newExchanges });
```

No schema migration needed — `line_start` and `line_end` are already stored per exchange.

## Testing

Verified locally: after `rebuild`, subsequent index runs correctly pick up new exchanges appended to previously-indexed files.

### DIFF
--- a/dist/indexer.js
+++ b/dist/indexer.js
@@ -214,22 +214,26 @@ export async function indexUnprocessed(concurrency = 1, noSummaries = false) {
                 const projectArchive = path.join(ARCHIVE_DIR, project);
                 const archivePath = path.join(projectArchive, file);
                 const summaryPath = archivePath.replace('.jsonl', '-summary.txt');
-                // Check if already indexed in database
-                const alreadyIndexed = db.prepare('SELECT COUNT(*) as count FROM exchanges WHERE archive_path = ?')
-                    .get(archivePath);
-                if (alreadyIndexed.count > 0)
-                    continue;
+                // Check high-water mark: how many lines of this file are already indexed
+                const maxIndexedResult = db.prepare('SELECT COALESCE(MAX(line_end), 0) as max_line FROM exchanges WHERE archive_path = ?').get(archivePath);
+                const maxIndexedLine = maxIndexedResult.max_line;
                 // Ensure parent dirs exist for subagent files
                 fs.mkdirSync(path.dirname(archivePath), { recursive: true });
-                // Archive if needed
-                if (!fs.existsSync(archivePath)) {
+                // Archive: always update if file was partially indexed (it may have grown)
+                if (!fs.existsSync(archivePath) || maxIndexedLine > 0) {
                     fs.copyFileSync(sourcePath, archivePath);
                 }
                 // Parse and check
                 const exchanges = await parseConversation(sourcePath, project, archivePath);
                 if (exchanges.length === 0)
                     continue;
-                unprocessed.push({ project, file, sourcePath, archivePath, summaryPath, exchanges });
+                // Only include exchanges beyond the high-water mark
+                const newExchanges = maxIndexedLine > 0
+                    ? exchanges.filter(e => e.lineStart > maxIndexedLine)
+                    : exchanges;
+                if (newExchanges.length === 0)
+                    continue;
+                unprocessed.push({ project, file, sourcePath, archivePath, summaryPath, exchanges: newExchanges });
             }
         }
     } // end sourceDir loop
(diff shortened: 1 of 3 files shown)


obra:
Hi @anastasiiaanfimova — thanks for diagnosing this and proposing the fix. The bug + your high-water-mark approach were correct. Re-derived in fa02569 via TDD: red test that indexes 2 exchanges, appends 3 more, re-indexes, asserts 5 in the DB. Closing in favor of the merged version, but the credit is yours.

Original reporter @jamster — your detection script and the file-distribution stats were what made this actionable. Thank you.

Closes #84.

PR #86, opened 2026-04-30
fix: remove --prefer-offline from MCP wrapper npm install (#76)

### LINKED ISSUE #76: npm install --prefer-offline in mcp-server-wrapper.js fails with ETARGET on stale cache
## Summary

On fresh plugin install, `cli/mcp-server-wrapper.js` runs `npm install --prefer-offline --no-audit --no-fund` as a cold-start step. When the user's local npm metadata cache is stale, `--prefer-offline` makes npm refuse to hit the registry for updated package metadata, and the install fails with `ETARGET: No matching version found for better-sqlite3@^12.4.1`. The MCP server then never starts, and Claude Code reports `Failed to reconnect to plugin:episodic-memory:episodic-memory`.

Removing `--prefer-offline` makes the install succeed in ~13 seconds on the same machine with zero other changes.

## Environment
- Plugin version: `1.0.15`
- OS: macOS 15 (Darwin 24.6.0, arm64)
- Node: v25.8.2
- npm: 11.11.1
- Claude Code plugin install via `/plugin install episodic-memory@superpowers-marketplace`

## Reproduction

1. Fresh plugin install (no existing `node_modules` in `~/.claude/plugins/cache/superpowers-marketplace/episodic-memory/1.0.15/`).
2. `/reload-plugins` in Claude Code — wrapper attempts auto-install.
3. Claude reports `Failed to reconnect to plugin:episodic-memory:episodic-memory`.
4. Run the wrapper's exact install command manually:
   ```
   cd ~/.claude/plugins/cache/superpowers-marketplace/episodic-memory/1.0.15
   npm install --prefer-offline --no-audit --no-fund
   ```
   Result:
   ```
   npm error code ETARGET
   npm error notarget No matching version found for better-sqlite3@^12.4.1.
   ```
5. Drop `--prefer-offline` and rerun:
   ```
   npm install --no-audit --no-fund
   ```
   Result:
   ```
   added 236 packages in 13s
   ```
   `better-sqlite3` postinstall native rebuild succeeds on Node v25.8.2 with no intervention.

`npm view better-sqlite3 versions` confirms `12.4.1` through `12.8.0` are all published and available — this is purely a local cache-freshness issue created by the flag.

## Root cause

`cli/mcp-server-wrapper.js:28`:

```js
const child = spawn(npmCommand, ['install', '--prefer-offline', '--no-audit', '--no-fund'], {
```

`--prefer-offline` tells npm to use its local metadata cache without revalidating against the registry. If the cache snapshot predates the minimum version pinned in `package.json`, the resolver can't satisfy the range and produces `ETARGET`. The flag is a trade of correctness for a small speed gain, and the speed gain doesn't matter here because `runNpmInstall()` only runs once in the plugin's lifetime (the `existsSync(nodeModulesPath)` check short-circuits all subsequent launches).

## Suggested fix

```diff
- const child = spawn(npmCommand, ['install', '--prefer-offline', '--no-audit', '--no-fund'], {
+ const child = spawn(npmCommand, ['install', '--no-audit', '--no-fund'], {
```

One-line change. Happy to open a PR if helpful.

## Side note

While debugging, I also verified that `better-sqlite3`'s postinstall native build works cleanly on Node v25.8.2 — no prebuilt binary yet, but it compiles from source in a few seconds with zero intervention. The "use Node v22/v24 LTS" guidance from #19 may be out of date for the current `better-sqlite3` release.

## Workaround for other users

Until fixed, anyone seeing `Failed to reconnect to plugin:episodic-memory:episodic-memory` on fresh install can manually run:

```
cd ~/.claude/plugins/cache/superpowers-marketplace/episodic-memory/<version>
npm install --no-audit --no-fund
```

then `/reload-plugins` in Claude Code.

mvanhorn:
Fixes #76.

## What changed

Removed `--prefer-offline` from the `npm install` invocation in `cli/mcp-server-wrapper.js`. The `--no-audit` and `--no-fund` flags are kept.

## Why this matters

`--prefer-offline` tells npm to skip the registry whenever local cache metadata is present, even if that metadata is stale. The plugin's `better-sqlite3` dependency range is `^12.4.1`. On a fresh install where the user's npm cache predates that range, `--prefer-offline` makes npm refuse to refresh, and the install fails with:

```
npm error code ETARGET
npm error notarget No matching version found for better-sqlite3@^12.4.1.
```

The MCP server never starts and Claude Code surfaces `Failed to reconnect to plugin:episodic-memory:episodic-memory`. The reporter (DarkbyteAT in #76) confirmed that simply removing the flag makes the install succeed in ~13 seconds with no other changes.

## Why drop it rather than work around it

The flag's only benefit is shaving registry round-trips on a path that already only runs once (when `node_modules/` does not exist). Trading a one-time install latency improvement for a hard failure mode on stale caches is not worth it for a wrapper that gates the entire plugin from starting.

## Verification

- `node -c cli/mcp-server-wrapper.js` parses cleanly.
- Diff is one line; no behavior change beyond removing the flag.

### DIFF
--- a/cli/mcp-server-wrapper.js
+++ b/cli/mcp-server-wrapper.js
@@ -25,7 +25,7 @@ function runNpmInstall() {
     console.error('This may take 30-60 seconds...');
 
     // Install dependencies - npm will auto-install optionalDependencies for current platform
-    const child = spawn(npmCommand, ['install', '--prefer-offline', '--no-audit', '--no-fund'], {
+    const child = spawn(npmCommand, ['install', '--no-audit', '--no-fund'], {
       cwd: PLUGIN_ROOT,
       stdio: ['ignore', 'pipe', 'pipe'],
       shell: isWindows // On Windows, we need shell: true to find npm.cmd


obra:
Thanks @mvanhorn! Merged. The reasoning in the PR body — that the once-per-install path doesn't gain meaningfully from --prefer-offline while losing correctness on stale caches — is exactly right. Native deps already require network for prebuild/registry on first install, so this isn't giving up real offline-install capability.

PR #93, opened 2026-05-18
fix: sync queue stalls on short sessions from other project directories

minyek:
# Sync queue stalls on short sessions from other project directories

## Description

In a normal SessionStart-triggered sync, short sessions (≤15 exchanges) archived from any project other than the one the sync process is running in silently fail to summarize. The sync's catch writes no `*-summary.txt`, so failed files re-queue every run. With `summaryLimit: 10`, once ten such failures accumulate at the head of the queue, real conversations behind them are never reached — same stall pattern as the zero-exchange zombie fix in 1.4.1, different trigger.

Log fingerprint:

```
✅ Sync complete!
  Summarized: 0
⚠️  Errors: 4
  <archive>/<project>/<uuid>.jsonl:
    Summary generation failed: Cannot read properties of undefined (reading 'match')
  ...
💡 All N summarization attempts failed.
  Check your API configuration (EPISODIC_MEMORY_API_BASE_URL / ANTHROPIC_API_KEY).
```

Two things are misleading: the API is fine, and `Cannot read properties of undefined` is a *secondary* crash. The actual SDK error (`No conversation found with session ID: ...`) is swallowed before it can propagate to the user-facing log.

This wasn't visible to most users on ≤1.4.0 because zero-exchange zombies were absorbing the queue head first. Now that 1.4.1 drains those, cwd-mismatch failures are the new front of the queue — effectively a regression that 1.4.1 exposes.

Mitigating factor: only sessions ≤15 exchanges take the resume path. Longer sessions include the conversation text directly and don't resume, so they're unaffected.

Distinct from the zero-exchange zombie fingerprint:

- `Summarizing <file> (K exchanges)...` lines appear (parser produced exchanges),
- followed by `Summarized: 0` and a non-empty `⚠️ Errors` block,
- and the same files re-appear in every subsequent sync.

## Environment

- episodic-memory v1.4.1
- `@anthropic-ai/claude-agent-sdk` bundled with v1.4.1

## Steps to Reproduce

1. Have any short session (≤15 exchanges) archived from project A.
2. Start a Claude Code session somewhere whose cwd differs from project A — SessionStart fires `episodic-memory sync --background` with that cwd.
3. Sync picks up the session-from-A file and tries to summarize it.
4. Observe `⚠️ Errors: N` with `Cannot read properties of undefined (reading 'match')`.
5. Re-run sync — same files, same errors, indefinitely.

## Root Cause

For short conversations (≤15 exchanges), `summarizeConversation` calls the Claude Agent SDK with `resume: sessionId` and *omits* the transcript text from the prompt — the resumed session is expected to provide context. Three coordinated issues then cascade:

**1. `buildSummarizerQueryOptions` does not set `cwd`.** The SDK spawns `claude` as a subprocess, which inherits the sync process's cwd. Claude Code resolves `resume: sessionId` against `~/.claude/projects/<encode(process.cwd())>/<sessionId>.jsonl`. If the session belongs to a different project, that file doesn't exist and the subprocess emits `type: 'result'` with `is_error: true` and `subtype: 'error_during_execution'`.

Verified directly with the bundled SDK against an archived session file. Same session ID, same options, same prompt — only cwd differs:

| cwd used to spawn | Result |
|---|---|
| Sync's cwd (≠ session's project) | `*** stream threw: No conversation found with session ID: ...` |
| The session's recorded project cwd | 9 messages streamed; real `<summary>` returned in ~5 s |

**2. `callClaude` swallows the SDK's error signal.** The current code matches `message.type === 'result'`, reads `message.result` (undefined on `is_error` results), and returns immediately. The SDK's iterator would have thrown a human-readable `"Claude Code returned an error result: No conversation found with session ID: ..."` on the next iteration, but the early return skips that. `callClaude` returns `undefined`, `extractSummary(undefined)` crashes with `Cannot read properties of undefined (reading 'match')`, and the actionable SDK error is unreachable.

**3. `sync.ts` writes no sentinel on failure.** The catch pushes to `result.errors` without creating a `*-summary.txt`, so the file remains in `filesToSummarize` on every subsequent sync — same forever-failure pattern as the zero-exchange zombie, different trigger.

## Why These Sessions Exist

cwd mismatch is the normal state for any multi-project install:

- **Active-project sessions.** The session's cwd still exists on disk; only the sync's cwd is wrong. Fully resumable when spawned with the right cwd.
- **Dead-worktree sessions.** The session's cwd no longer exists (worktree deleted, branch removed). Even chdir-ing can't recover these.
- **Multi-project sessions.** Sessions whose cwd-encoded paths collide ambiguously after archive copy.

Sessions outlive the cwds they came from — that's correct archive behavior. The bug is the summarizer's failure to handle the mismatch.

## Fix Summary

Three parts; see the diff for code.

1. **Pass the session's recorded `cwd` to the SDK** when the directory still exists on disk. The parser already exposes `cwd` on each exchange. Resolves the active-project case.
2. **Make `callClaude` throw `SummarizerSdkError` on SDK `is_error` results** instead of silently returning `undefined`. Prerequisite for (3) — without this, the resume failure is invisible to any catch-and-retry logic, and the only error to propagate is the secondary `Cannot read properties of undefined` from `extractSummary`, which doesn't pattern-match any resume signal.
3. **Fall back to non-resume on `isResumeFailure`** — retry once without `resume:` and include the conversation text in the prompt. Resolves the dead-worktree case where no cwd on disk can ever satisfy the SDK.

Tests added: `test/summarizer-options.test.ts` (pure-function coverage for the new `cwd` plumbing and `isResumeFailure`) and `test/summarizer-resume-fallback.test.ts` (integration coverage via a mocked SDK — error throw, fallback path with conversation text included, cwd-existence guards, non-resume errors propagating without firing the fallback).

## Fix Verified

Applied to a real install with 4 cwd-mismatch files pinning the queue (1 active project, 3 dead worktrees). Drain log:

```
Generating summaries for 4 conversation(s)...
  Summarizing <uuid-1>.jsonl (2 exchanges)...     [active project — Part 1]
  Summarizing <uuid-2>.jsonl (1 exchanges)...     [dead worktree — Part 3]
  Summarizing <uuid-3>.jsonl (8 exchanges)...     [dead worktree — Part 3]
  Summarizing <uuid-4>.jsonl (13 exchanges)...    [dead worktree — Part 3]
    resume failed for <uuid-4> (SummarizerSdkError: error_during_execution); retrying without resume

✅ Sync complete!
  Summarized: 4
```

All four cleared. The next sync had no `Generating summaries` line — queue fully drained.

Caveat: some fallback summaries return raw model output rather than `<summary></summary>`-wrapped text. The prompt was originally written for the resume case (where the system prompt belongs to the resumed session); when reused for the fallback, the model sometimes omits the wrapper. `extractSummary` falls back to returning the raw text — searchable and sentineled, but worth tightening if upstream wants summary uniformity.

### DIFF
--- a/dist/summarizer.d.ts
+++ b/dist/summarizer.d.ts
@@ -1,4 +1,19 @@
 import { ConversationExchange } from './types.js';
+/**
+ * Thrown by callClaude when the SDK yields an `is_error: true` result message.
+ * Carries the SDK's `subtype` and `session_id` as typed fields so callers can
+ * dispatch on structural metadata rather than parsing error message text.
+ */
+export declare class SummarizerSdkError extends Error {
+    readonly subtype: string;
+    readonly sessionId?: string | undefined;
+    constructor(subtype: string, sessionId?: string | undefined);
+}
+/**
+ * True when the SDK's reported failure subtype indicates resume couldn't find
+ * the session — the trigger for the non-resume fallback in summarizeConversation.
+ */
+export declare function isResumeFailure(error: unknown): boolean;
 export interface CodexSummarizerCommand {
     command: string;
     args: string[];
@@ -40,6 +55,7 @@ export declare function formatConversationText(exchanges: ConversationExchange[]
 export declare function buildSummarizerQueryOptions(args: {
     model: string;
     sessionId?: string;
+    cwd?: string;
 }): Record<string, unknown>;
 export declare function buildCodexSummaryPrompt(): string;
 export declare function buildCodexSummarizerCommand(args: {
--- a/dist/summarizer.js
+++ b/dist/summarizer.js
@@ -1,9 +1,32 @@
+import fs from 'fs';
 import { query } from '@anthropic-ai/claude-agent-sdk';
 import { SUMMARIZER_CONTEXT_MARKER } from './constants.js';
 import { VERSION } from './version.js';
 import { spawn } from 'child_process';
 import { createInterface } from 'readline';
 import { codexVersionRequirementMessage, parseCodexCliVersion, versionMeetsMinimum, } from './codex-support.js';
+/**
+ * Thrown by callClaude when the SDK yields an `is_error: true` result message.
+ * Carries the SDK's `subtype` and `session_id` as typed fields so callers can
+ * dispatch on structural metadata rather than parsing error message text.
+ */
+export class SummarizerSdkError extends Error {
+    subtype;
+    sessionId;
+    constructor(subtype, sessionId) {
+        super(`Summarizer SDK error: ${subtype}${sessionId ? ` (session ${sessionId})` : ''}`);
+        this.subtype = subtype;
+        this.sessionId = sessionId;
+        this.name = 'SummarizerSdkError';
+    }
+}
+/**
+ * True when the SDK's reported failure subtype indicates resume couldn't find
+ * the session — the trigger for the non-resume fallback in summarizeConversation.
+ */
+export function isResumeFailure(error) {
+    return error instanceof SummarizerSdkError && error.subtype === 'error_during_execution';
+}
 /**
  * Get API environment overrides for summarization calls.
  * Returns full env merged with process.env so subprocess inherits PATH, HOME, etc.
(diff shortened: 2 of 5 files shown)


obra:
Thanks @minyek! Merged manually as squash commit 4e89881 with your authorship preserved.

The PR's commits had a trivial import-list conflict with #99 (both PRs added new exports to test/summarizer-options.test.ts). Resolved locally by keeping both sets of imports — no functional change to your code.

I traced the data path end-to-end and the fix is exactly right:
- The parser already extracts `payload.cwd` from JSONL into `ConversationExchange.cwd` (src/parser.ts:480-489), so `exchanges.find(e => e.cwd)?.cwd` picks it up cleanly.
- Real archived JSONL files in the wild all carry the field.
- All 175 tests pass (158 baseline + your 13 new + #99's 4 new) post-rebase.

Three-part fix is well-motivated; the `is_error` throw being load-bearing for the fallback ("without it the catch sees only the secondary extractSummary crash, which no resume-failure pattern matches") is exactly the kind of thing that's easy to miss when refactoring this code later — thanks for being explicit about it in the commit message.

Also tidied up a pre-existing main drift (dist/version.* was committed as 1.4.0 since the v1.4.1 release didn't regenerate it).

— Claude Opus 4.7, Claude Code 2.1.146

PR #99, opened 2026-05-21
Fix Codex summarizer: ignore deprecated model from history; add env override

### LINKED ISSUE #98: Codex summarizer: hardcoded model from history breaks on deprecated model names and ChatGPT-account users
## Problem

`getCodexModel()` in `src/summarizer.ts` reads the model from the historical exchange data:

```ts
function getCodexModel(exchanges: ConversationExchange[]): string | undefined {
  return exchanges.find(exchange => exchange.harness === 'codex' && exchange.model)?.model;
}
```

That value gets passed verbatim into `thread/fork` via `app-server`. For older conversations this returns a model name that no longer works, and the entire `Summarized` count drops to 0 for Codex transcripts.

## Two failure modes I hit on macOS (codex-cli 0.132.0, ChatGPT subscription auth)

1. **Deprecated model name** — Codex conversations from earlier in 2026 have `model: "gpt-5.2-codex"` baked in. The fork call returns:
   ```
   {"type":"error","status":400,"error":{"type":"invalid_request_error",
    "message":"The 'gpt-5.2-codex' model is not supported when using Codex with a ChatGPT account."}}
   ```

2. **Any `-codex` suffix variant rejected for ChatGPT-subscription users** — I patched `getCodexModel` to force `gpt-5.5-codex`. Same 400, identical wording. Replacing with plain `gpt-5.5` then succeeds. So the error message is misleading: the `-codex` variants are API-only; ChatGPT subscriptions can only use the base models (`gpt-5.5`, etc.) for `app-server` fork.

End result: ~250 Codex transcripts in my archive fall through to `Codex summarizer unavailable, falling back to transcript text`. Their fallback summary is the raw transcript, which hurts recall quality.

## Repro

1. Have at least one Codex session whose exchanges carry `model: "gpt-5.2-codex"` (any pre-deprecation conversation).
2. Run `episodic-memory sync` on a ChatGPT-subscription Codex login.
3. Every Codex transcript fails with the 400 above.

## Suggested fix

The fix that worked for me was making `getCodexModel` ignore the historical value and let Codex pick the default from `~/.codex/config.toml`:

```ts
function getCodexModel(exchanges: ConversationExchange[]): string | undefined {
  if (process.env.EPISODIC_MEMORY_CODEX_MODEL) {
    return process.env.EPISODIC_MEMORY_CODEX_MODEL;
  }
  return undefined;  // let codex use config.toml default
}
```

With `model = "gpt-5.5"` in my `~/.codex/config.toml`, this round-trips correctly and the 247-conversation backlog drains cleanly.

Two reasons this is better than reading from history:
- Codex model ids drift (the `gpt-5.2-codex` deprecation will keep happening), but `config.toml` always points at something current.
- ChatGPT-subscription users get a working default; API-key users can still set `EPISODIC_MEMORY_CODEX_MODEL=gpt-5.5-codex` if they want the codex-tuned variant.

Happy to send a PR if you want this exact shape.

## Environment

- macOS 15.5 (darwin arm64)
- codex-cli 0.132.0
- episodic-memory 1.4.1
- ChatGPT subscription auth (no OpenAI API key)

monsterxz9:
## Summary

- `getCodexModel()` now defaults to `undefined`, letting Codex's `app-server` fork pick the model from `~/.codex/config.toml`
- Operators who need a specific model id (e.g. API-key users wanting `gpt-5.5-codex`) can set `EPISODIC_MEMORY_CODEX_MODEL`
- Added four unit tests covering the default + env-override paths

## Why

Reading the model from historical exchanges breaks Codex summarization for two common populations — see #98 for the full repro. tl;dr:

1. Pre-deprecation Codex sessions carry `model: "gpt-5.2-codex"`, which `app-server` now rejects with `400 "model is not supported when using Codex with a ChatGPT account"`.
2. Any `-codex`-suffix variant (`gpt-5.5-codex`, etc.) is API-key-only — ChatGPT-subscription users get the **same** 400 regardless of which suffix is used. The error message is misleading; only base models (`gpt-5.5`, ...) are reachable via `app-server` fork for them.

In practice this dropped ~250 of my Codex transcripts into the `Codex summarizer unavailable, falling back to transcript text` path. After applying this change, the backlog drained cleanly on the next sync.

## Compat / behavior

- Default behavior changes: previously the historical model was passed to `thread/fork`; now nothing is passed unless `EPISODIC_MEMORY_CODEX_MODEL` is set. Codex falls back to `config.toml#model`, which is what `codex` itself uses interactively — same model the user sees in their UI.
- API-key users who were relying on the historical `-codex` variant being forwarded can restore that behavior with `EPISODIC_MEMORY_CODEX_MODEL=gpt-5.5-codex` (or whichever variant they want).

## Test plan

- [x] `npm test test/summarizer-options.test.ts` — 19/19 pass (4 new tests for `getCodexModel`)
- [x] `npm run build` — clean
- [x] Manual repro: ran `EPISODIC_MEMORY_CODEX_MODEL=gpt-5.5 episodic-memory sync` against an archive of 247 Codex transcripts originally summarized as `gpt-5.2-codex`. All summaries succeeded; backlog drained.
- Pre-existing failures on `main` in `test/show.test.ts` (hardcoded date `2025-09-19`) and `test/multi-concept.test.ts` (probabilistic ranking) are unrelated to this change.

Closes #98.

### DIFF
--- a/dist/mcp-server.js
+++ b/dist/mcp-server.js
@@ -26850,7 +26850,7 @@ ${result}
 }
 
 // src/version.ts
-var VERSION = "1.4.0";
+var VERSION = "1.4.1";
 
 // src/mcp-server.ts
 import fs4 from "fs";
--- a/dist/summarizer.d.ts
+++ b/dist/summarizer.d.ts
@@ -49,4 +49,20 @@ export declare function buildCodexSummarizerCommand(args: {
     codexBin?: string;
 }): CodexSummarizerCommand;
 export declare function runCodexCommand(command: CodexSummarizerCommand): Promise<string>;
+/**
+ * Resolve the model to pass into Codex `thread/fork` for summarization.
+ *
+ * Historical exchanges may carry deprecated model ids (e.g. `gpt-5.2-codex`),
+ * and `-codex`-suffixed variants are API-key-only — ChatGPT-subscription users
+ * get a 400 from `app-server` regardless of the suffix used. Reading the model
+ * from history therefore breaks summarization for two large user populations.
+ *
+ * Default to `undefined` so `app-server` uses the current Codex config
+ * (`~/.codex/config.toml#model`). Operators can override via
+ * `EPISODIC_MEMORY_CODEX_MODEL` if they need a specific model id (e.g. an
+ * API-key user wanting `gpt-5.5-codex`).
+ *
+ * See https://github.com/obra/episodic-memory/issues/98.
+ */
+export declare function getCodexModel(_exchanges: ConversationExchange[]): string | undefined;
 export declare function summarizeConversation(exchanges: ConversationExchange[], sessionId?: string): Promise<string>;
--- a/dist/summarizer.js
+++ b/dist/summarizer.js
@@ -347,8 +347,23 @@ function getCodexSessionId(exchanges, sessionId) {
     }
     return sessionId || exchanges.find(exchange => exchange.sessionId)?.sessionId;
 }
-function getCodexModel(exchanges) {
-    return exchanges.find(exchange => exchange.harness === 'codex' && exchange.model)?.model;
+/**
+ * Resolve the model to pass into Codex `thread/fork` for summarization.
+ *
+ * Historical exchanges may carry deprecated model ids (e.g. `gpt-5.2-codex`),
+ * and `-codex`-suffixed variants are API-key-only — ChatGPT-subscription users
+ * get a 400 from `app-server` regardless of the suffix used. Reading the model
+ * from history therefore breaks summarization for two large user populations.
+ *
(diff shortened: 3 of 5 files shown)


obra:
Thanks @monsterxz9! Merged. I verified the fix end-to-end against a real ChatGPT-account Codex install (codex-cli 0.130.0): `gpt-5.2-codex` and `gpt-5.5-codex` both 400 with the exact error text you quoted, and omitting `model` from `thread/fork` cleanly falls through to `~/.codex/config.toml#model`. Pure-function tests pass, full suite (162/162) clean.

— Claude Opus 4.7, Claude Code 2.1.146

PR #101, opened 2026-05-25
Enable busy_timeout + foreign_keys for multi-writer safety

Dmheath1:
Multiple Claude Code sessions concurrently writing to the same SQLite store produces corruption. better-sqlite3 doesn't enable WAL by default (though this repo already set it). The PRAGMAs added here are standard for any multi-process SQLite:

- \`busy_timeout = 5000\` — wait up to 5s for a lock before failing
- \`foreign_keys = ON\` — restore FK enforcement after the sqlite-vec extension load path disables it

Applied to all three Database connection points: \`src/db.ts\` (main writer), \`dist/mcp-server.js\` (separate MCP-server writer that wasn't covered by the existing sync-cli lock), and \`src/stats.ts\` (readonly connection that can still hit busy during WAL checkpoint).

Includes a vitest concurrent-read-during-write smoke test.

### DIFF
--- a/dist/db.js
+++ b/dist/db.js
@@ -102,6 +102,8 @@ export function initDatabase() {
     sqliteVec.load(db);
     // Enable WAL mode for better concurrency
     db.pragma('journal_mode = WAL');
+    db.pragma('busy_timeout = 5000');
+    db.pragma('foreign_keys = ON');
     // Create exchanges table
     db.exec(`
     CREATE TABLE IF NOT EXISTS exchanges (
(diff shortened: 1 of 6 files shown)


obra:
Thanks for the contribution. Would you mind resubmitting this without a change to the version number inside the mcp?

Dmheath1:
Done. VERSION revert in a470b5f. PR diff is now just the two PRAGMA lines plus the smoke test.

obra:
Thanks @Dmheath1! Merged to main as `f5e85f3` (squashed as part of a 13-PR batch, with dist rebuilt on top). This adds a busy_timeout for safer multi-writer access to the SQLite index.

Issue #87, opened 2026-05-03
Version 1.1.0 spawns cc session


kaankoken:
As soon as I launched a CC session, my Mac's memory hit max. The latest version spawns a zombie cc session, and i am not able use my computer. 

Several times, I had to shut down the computer manually.

- CC: 2.1.126
- macOS: 26.4
- plugin version: 1.1.0

I downgraded the plugin to fix my issue for now.

kaankoken:
<img width="1676" height="1106" alt="Image" src="https://github.com/user-attachments/assets/5df70e30-be35-424a-b4f5-2e971ac1d748" />

@obra

materemias:
Adding observations from a local incident that looks like the same root cause. I originally opened #88, then closed it as duplicate of this.

What I observed live:

- Many detached `episodic-memory/1.1.0/dist/sync-cli.js` processes under `systemd --user`.
- Each sync process was spawning Claude Agent SDK children like `.../claude-agent-sdk-linux-x64/claude --model haiku ...`.
- Those child Claude sessions loaded normal Claude Code hooks/MCP config.
- Because the plugin has a `SessionStart` hook that runs `episodic-memory sync --background`, those SDK-launched Claude sessions could trigger another background sync.
- CPU and token usage exploded quickly. I also had GitKraken/GitLens AI hooks installed (`gk ai hook run --host claude-code`), which amplified the session/event fanout, but the recursive edge appears to be: `SessionStart sync --background` -> summarizer uses Claude Agent SDK -> spawned Claude triggers `SessionStart` again.

Relevant installed hook:

```json
{
  "hooks": {
    "SessionStart": [
      {
        "matcher": "startup|resume|clear",
        "hooks": [
          {
            "type": "command",
            "command": "node \"${CLAUDE_PLUGIN_ROOT}/cli/episodic-memory.js\" sync --background",
            "async": true
          }
        ]
      }
    ]
  }
}
```

Relevant sync/summarization behavior from 1.1.0:

- `sync.ts` collects conversations without summaries and summarizes up to 10 per sync.
- `summarizer.ts` calls `@anthropic-ai/claude-agent-sdk` `query()` with model `haiku` by default.
- The summarizer options include `persistSession: false`, but that did not prevent the child Claude process from loading hooks/config in this environment.

Impact:

- CPU saturated.
- Hundreds of Claude/MCP/GitKraken-hook-related processes appeared.
- Claude quota/session tokens were consumed very quickly.
[...]

PR #108, opened 2026-06-07
fix: run Claude summarization tool-less so it can't hijack an active session

minyek:
Background summarization could hijack an active Claude session and run commands in the user's working tree. To summarize a recent conversation the plugin **resumes that Claude session** and asks it for a `<summary>` — but it did so with full default tool access. When the resumed session was still **mid-task**, the model inherited the task's momentum *and* the ability to execute tools, so it kept *executing* the original task instead of summarizing.

Process fingerprint — a second `claude` parented under `sync-cli.js`, resuming a live session with full permissions:

```
…/claude-agent-sdk-linux-x64/claude --output-format stream-json --verbose \
  --input-format stream-json --model haiku --resume c7691203-9342-… \
  --permission-mode default --no-session-persistence
```

Observed during an active worktree session: the spawned agent committed a work-in-progress refactor on the user's branch with a model-generated message, along with various other tool uses. It recurred every sync cycle (~2–2.5h) and had to be killed each time.

## Root cause

`buildSummarizerQueryOptions()` in `src/summarizer.ts` set no tool, permission, or sandbox restriction, so the resumed Claude ran under `--permission-mode default` (full tool access). Summarization is text-only work, but nothing stopped the model from invoking `Bash`/`Edit`/`Write`/`Task`. This is asymmetric with the **Codex** summarizer path in the same file, which already forks read-only (`sandbox: 'read-only'`, `approvalPolicy: 'never'`). The `#87` reentrancy guard is orthogonal — it stops the recursive sync→summarize→sync cascade but does not restrict tool execution.

## Fix

Restrict the Claude summarization query to be tool-less — the Claude-side equivalent of the Codex `sandbox: 'read-only'` — via `tools: []` in `buildSummarizerQueryOptions()`. Per the Agent SDK `Options` docs, `tools` is the documented way to restrict which tools are available, and `[]` disables every built-in, so a resumed mid-task session has no tool to invoke and can only emit text.

(`allowedTools` is only an auto-approve list, not a restriction, and a `disallowedTools` denylist would be a strict subset of what `tools: []` already removes — so neither adds protection over `tools: []`.)

## Testing

New assertions in `test/summarizer-options.test.ts` pin the restriction so a future refactor can't silently drop it: `tools: []` verified on both the resume and fresh-session paths.

Also includes a fix to `test/show.test.ts`: its timestamp assertion used a hardcoded US/ISO regex and failed under non-US locales; it now computes the expected string via `toLocaleString()` the same way `show.ts` does, making it locale/timezone-independent.

## Verification

`npx vitest run test/summarizer-options.test.ts test/show.test.ts` → 44 passed. `dist/` rebuilt from `src/` and committed alongside it.

## Scope

This PR implements the must-have tool restriction. A separate, related issue — a session summarized mid-flight has its summary frozen at the partial state and never re-summarized as it grows — is tracked and fixed in a follow-up PR (#109) . The tool restriction alone defangs the runaway; the follow-up closes the frozen-partial gap.

### DIFF
--- a/dist/summarizer.d.ts
+++ b/dist/summarizer.d.ts
@@ -51,6 +51,9 @@ export declare function formatConversationText(exchanges: ConversationExchange[]
  * ~/.claude/projects/ (#83). Without it, every summarization spawns a fake
  * session JSONL that pollutes the IDE session sidebar. The option is honored
  * by claude-agent-sdk >= 0.2.0.
+ *
+ * tools: [] disables every built-in tool so a resumed mid-task session can't
+ * keep EXECUTING the task instead of writing a <summary>.
  */
 export declare function buildSummarizerQueryOptions(args: {
     model: string;
--- a/dist/summarizer.js
+++ b/dist/summarizer.js
@@ -85,6 +85,9 @@ function extractSummary(text) {
  * ~/.claude/projects/ (#83). Without it, every summarization spawns a fake
  * session JSONL that pollutes the IDE session sidebar. The option is honored
  * by claude-agent-sdk >= 0.2.0.
+ *
+ * tools: [] disables every built-in tool so a resumed mid-task session can't
+ * keep EXECUTING the task instead of writing a <summary>.
  */
 export function buildSummarizerQueryOptions(args) {
     const { model, sessionId, cwd } = args;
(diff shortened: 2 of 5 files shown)


obra:
Thanks @minyek! Merged to main as `41e135a` (squashed as part of a 13-PR batch, with dist rebuilt on top). This runs summarization tool-less so a resumed session can't execute its pending work.

Issue #96, opened 2026-05-19
Failed summaries are re-queued forever, inflating backlog (v1.4.1)


bradfeld:
> _This issue was drafted by Claude and reviewed by Brad Feld before posting._

## Summary

In v1.4.1, files where `summarizeConversation()` fails are re-queued on every subsequent sync run, indefinitely. Combined with a missing null-guard in `extractSummary()`, this produces a permanent backlog inflation: errored files always re-fail, the backlog never drains, and each failure emits an "API Error - session interrupted" macOS notification from the spawned `claude --model haiku` subprocess.

In my environment, the backlog went from 789 → 768 across 11 sync runs (going *up*, not down), with 80 `Cannot read properties of undefined (reading 'match')` errors in the log.

### 1. `dist/summarizer.js` line 49: missing null guard

```js
function extractSummary(text) {
    const match = text.match(/<summary>(.*?)<\/summary>/s);
    ...
}
```

When `callClaude()` returns `undefined` (e.g., the Claude Agent SDK subprocess crashes mid-call, or any transient failure prevents text capture), `text` is `undefined` and `text.match(...)` throws `TypeError: Cannot read properties of undefined (reading 'match')`. This is the proximate cause of the error noise.

### 2. `dist/sync.js` line 168: no sentinel on error

```js
catch (error) {
    result.errors.push({
        file: filePath,
        error: `Summary generation failed: ${...}`
    });
}
```

The empty-conversation branch at line 156 correctly writes an empty `-summary.txt` sentinel so trivial files aren't re-queued. The error branch does not. As a result, any file whose summarization fails (whether transient or persistent) gets re-attempted on every sync run forever, with the same failure each time.

## Reproduction

1. Have a large backlog of unsummarized conversations (~800 in my case)
2. Run `node dist/sync-cli.js` repeatedly
3. Watch `(N more need summaries - will process on next sync)` — N stays flat or grows
4. Check the error log: persistent `Cannot read properties of undefined (reading 'match')` for the same files run after run

## Suggested fix

Two-line change in each file. I applied these locally and they resolve both symptoms (`match` errors stop; sentinels written for failed files; previously-re-queued files no longer come back).

**`summarizer.js`:**
```js
function extractSummary(text) {
    if (typeof text !== 'string' || text.length === 0) {
        throw new Error('Summarizer returned no text (subprocess likely failed)');
    }
    const match = text.match(/<summary>(.*?)<\/summary>/s);
    ...
}
```

[...]

PR #109, opened 2026-06-07
fix: summarize once idle, refresh on growth, and bound failed-summary retries

minyek:
## Description

Summaries are **write-once**. A conversation that gets summarized while still in
progress keeps a summary of only its early portion **forever** — every later turn
is invisible to the summary. For long-running or resumed conversations, the
one-line summary shown next to search results is therefore misleading.

The fix has two complementary halves: **defer** summarizing a conversation until
it has gone idle (so a still-active conversation is never frozen into a partial
snapshot), and **refresh** a summary when its transcript later grows. Note the
behavioral change from the defer half: a conversation is no longer summarized
until it has been idle for the quiescence window (`EPISODIC_MEMORY_SUMMARY_QUIESCENCE_HOURS`,
default 1h), so very recent conversations intentionally have no summary yet.

Because the refresh half makes us re-summarize, failures can now *recur* — so the
same change also makes failed (re-)summaries **back off and eventually give up**
instead of preserving a good summary but re-attempting on every sync. That retry
policy is unified with the existing first-failure error-sentinel path (see Fix 5).

(Search results themselves were always complete — search runs over indexed
*exchanges*, kept fresh via `INSERT OR REPLACE`. This fixes only the summary
displayed alongside them.)

## Symptom

There is no error in the logs — the symptom is silently wrong output, plus
silently wasted work:

- `<archive>/<project>/<session>-summary.txt` for an active or long conversation
  describes only the first handful of exchanges.
- Continuing the conversation over hours/days never updates it.
- `stats`/`verify` count the file as "summarized", so it is never revisited.
- A conversation that *can't* be summarized (oversized transcript, a turn that
  always errors) re-invokes the summarizer on every sync, indefinitely.

## Environment

- episodic-memory v1.4.2
- Claude Agent SDK summarizer
- All installs affected; most visible to anyone who keeps long or resumed
  sessions.

## Steps to Reproduce

1. Start a conversation and let a sync summarize it early (a few exchanges).
2. Continue the conversation for many more turns over time.
3. Run `sync` again (or let the SessionStart background sync run).
4. Inspect `<archive>/<project>/<session>-summary.txt`.
5. Observe it still summarizes only the early portion — never re-summarized,
   even though the transcript has grown substantially.

## Root Cause

`shouldQueueForSummary` only re-queued a conversation when its sentinel was
**missing** or a **stale error marker**. Once a real summary existed, the
conversation was treated as done permanently:

1. **No coverage signal.** Nothing recorded *how much* of the transcript a given
   summary reflected, so transcript growth was undetectable.
2. **Append-only transcripts.** A summary written at exchange 5 is never
   refreshed at exchange 500 — the file only ever grows, but nothing acted on
   that growth.
3. **No idle awareness.** Even if growth had been detected, summarizing an
   in-progress conversation would freeze a mid-session snapshot that is itself
   immediately stale.

And once re-summarization exists, a fourth gap appears in failure handling:

4. **Unbounded retries.** A failed re-summary correctly kept the prior summary
   but recorded nothing, so the file re-queued and re-invoked the LLM on *every*
   sync forever. First-ever failures (`#96` error sentinels) backed off ~1h but
   never gave up. And since sync is `SessionStart`-driven, not a fixed timer, a
   1h floor rarely even skips a sync — the retry effectively fired every sync.

## Fix Summary

1. **Coverage header.** Real summaries now carry a machine-readable first line —
   `__COVERAGE__ {"bytes":<n>,"lastExchange":"<iso>","schema":1}` — recording the
   archived JSONL byte size the summary reflects (`formatSummaryFile` /
   `parseSummaryFile`, `buildCoverage`, `writeSummary`).
2. **Growth-driven re-queue.** `shouldQueueForSummary(path, currentArchiveBytes)`
   re-queues a real summary when the archive has grown past its covered bytes
   (append-only ⇒ a size increase means new content).
3. **Quiescence gate.** `isQuiescent` + `EPISODIC_MEMORY_SUMMARY_QUIESCENCE_HOURS`
   (default `1`) only (re)summarize after the conversation has been idle for the
   window, so a mid-session snapshot is never frozen. Idle is measured from the
   **last exchange's timestamp** — the archive copy's mtime is reset on every
   sync and is unreliable.
4. **Gentle upgrade.** Legacy header-less summaries are stamped with a
   current-size baseline (`ensureCoverageBaseline`) — a header rewrite, **no LLM
   call** — so existing installs gain a growth baseline without a mass
   re-summarization on upgrade.
5. **Preserve, back off, then give up on failure.** A failed (re-)summary keeps
   any prior real summary and records the attempt; one shared policy
   (`shouldRetryAfterFailure`) then holds the file until the retry floor
   (`EPISODIC_MEMORY_SUMMARY_ERROR_RETRY_HOURS`, default 1h) elapses and abandons
   it after `EPISODIC_MEMORY_SUMMARY_MAX_ATTEMPTS` (default 5) — keeping the last
   good summary rather than thrashing. This same policy governs **both** first-ever
   failures (`#96` error sentinels, now attempt-counted) and re-summary failures
   (state in the coverage header's `resummary` field), via `recordSummaryFailure`.
   It replaces the prior mtime-based, never-give-up retry and the duplicated
   floor/cap logic.
6. **Fair queue ordering.** Never-summarized conversations are processed ahead of
   re-summary retries before the per-run `summaryLimit`, so a backlog of failing
   re-summaries can't starve fresh work (the `#91` head-of-queue failure mode).
7. **Display.** `search` strips the coverage header from the summary shown
   alongside results.

Internally, the gate + write + failure logic is shared (`needsSummary`,
`recordSummaryFailure`, `shouldRetryAfterFailure`, `summarizeIfQuiescent`) across
the three indexer paths and the sync loop, and the `_HOURS` config defaults
convert to milliseconds at a single boundary (`MS_PER_HOUR`).

## Configuration

- `EPISODIC_MEMORY_SUMMARY_QUIESCENCE_HOURS` (default `1`) — idle time before a
  conversation is (re)summarized.
- `EPISODIC_MEMORY_SUMMARY_ERROR_RETRY_HOURS` (default `1`) — minimum wait between
  failed summary attempts (the retry floor).
- `EPISODIC_MEMORY_SUMMARY_MAX_ATTEMPTS` (default `5`) — failed attempts before
  giving up and keeping any existing summary. **(new)**

## Error-sentinel format change

The `#96` error sentinel's second line changes from a bare ISO timestamp to a
JSON failure-state line, so it can carry the attempt count:

```
before:  __ERRORED__\n2026-06-07T10:00:00.000Z\n<message>
after:   __ERRORED__\n{"attempts":2,"lastAttempt":1718040000000}\n<message>
```

Backward compatible: `parseErrorSentinel` reads the legacy v1.4.2 ISO form
(→ `attempts: 0`, `lastAttempt` from the timestamp), so sentinels already on disk
keep backing off correctly and are rewritten in the new form on their next
failure. An unrecognised second line degrades safely to `{attempts: 0,
lastAttempt: 0}` → retried, never lost.

## Tests added

- `test/summary-coverage.test.ts` — coverage header round-trip and legacy baseline stamping.
- `test/summary-quiescence.test.ts` — quiescence verdicts and `_HOURS` env parsing.
- `test/summary-sentinel.test.ts` — growth-aware gate decision table; unified
  backoff/give-up for both error sentinels and re-summaries; `parseErrorSentinel`
  round-trip + legacy ISO fallback; `getMaxSummaryAttempts`; `recordSummaryFailure`.
- `test/summary-queue-order.test.ts` — fresh-before-retry ordering (`#91` guard).
- `test/sync-resummary.test.ts` — growth-driven re-summary, no-op when unchanged,
  preserve-on-failure (now also asserts the attempt is recorded).
- `test/search-summary-strip.test.ts` — the header never leaks into the displayed summary.
- Updates to `sync-error-sentinel` (mtime-aging → stored-timestamp; sentinel-format
  assertion), `show`, and `sync` tests for the new header/gate.

## Fix Verified

- **Full suite:** 258 tests across 41 files pass.
- **Real end-to-end** on the built `dist` (isolated temp dirs, real summarizer):
  - Seed an 8 237-byte conversation → first sync writes a summary stamped `"bytes":8237`.
  - Grow the transcript to 41 069 bytes → next sync re-summarizes; coverage re-stamped `"bytes":41069`.
  - No further growth → next sync performs no re-summary.

## Caveat / Notes

- Re-summary is triggered by **byte growth**, not an exchange-count delta. This
  is deliberate: transcripts are append-only and the per-sync coverage scan is
  ~23 ms even against a multi-GB archive, so no recency cutoff or minimum-delta
  threshold is needed. The quiescence gate bounds work to one re-summary per
  idle-after-growth episode.
- Give-up is **permanent** for a given conversation (it keeps its last good
  summary, if any): the attempt cap bounds cost on a transcript that never
  summarizes. The cap is the real bound, since the coarse, variable sync cadence
  makes the wall-clock floor too short to reliably skip a sync; the floor's job
  is to stop rapid session restarts from spending the budget on a transient blip.

### DIFF
--- a/dist/indexer.js
+++ b/dist/indexer.js
@@ -5,7 +5,7 @@ import { parseConversation } from './parser.js';
 import { initEmbeddings, generateExchangeEmbedding } from './embeddings.js';
 import { summarizeConversation } from './summarizer.js';
 import { getArchiveDir, getExcludedProjects, getConversationSourceDirs, findJsonlFiles } from './paths.js';
-import { needsSummary, isQuiescent, writeSummary, writeErrorSentinelIfNew, } from './summary-sentinel.js';
+import { needsSummary, isQuiescent, writeSummary, recordSummaryFailure, } from './summary-sentinel.js';
 // Set max output tokens for Claude SDK (used by summarizer)
 process.env.CLAUDE_CODE_MAX_OUTPUT_TOKENS = '20000';
 // Increase max listeners for concurrent API calls
@@ -26,9 +26,10 @@ function sessionIdForSummary(exchanges) {
 }
 /**
  * Summarize one conversation once it has gone quiescent, writing the summary with
- * its coverage header. Preserves a prior summary on failure (an error sentinel is
- * written only for a first-time failure). Returns the summary, or null when the
- * conversation is skipped (not yet quiescent) or summarization fails.
+ * its coverage header. On failure, recordSummaryFailure preserves any prior
+ * summary and records the attempt (a first-time failure writes an error sentinel
+ * instead). Returns the summary, or null when the conversation is skipped (not yet
+ * quiescent) or summarization fails.
  */
 async function summarizeIfQuiescent(summaryPath, archivePath, exchanges, label) {
     if (!isQuiescent(exchanges, Date.now()))
@@ -40,7 +41,7 @@ async function summarizeIfQuiescent(summaryPath, archivePath, exchanges, label) 
         return summary;
     }
     catch (error) {
-        writeErrorSentinelIfNew(summaryPath, error);
+        recordSummaryFailure(summaryPath, error);
         console.log(`  ✗ ${label}: ${error}`);
         return null;
     }
--- a/dist/indexer.js
+++ b/dist/indexer.js
@@ -5,7 +5,7 @@ import { parseConversation } from './parser.js';
 import { initEmbeddings, generateExchangeEmbedding } from './embeddings.js';
 import { summarizeConversation } from './summarizer.js';
 import { getArchiveDir, getExcludedProjects, getConversationSourceDirs, findJsonlFiles } from './paths.js';
-import { formatErrorSentinel, shouldQueueForSummary } from './summary-sentinel.js';
+import { needsSummary, isQuiescent, writeSummary, writeErrorSentinelIfNew, } from './summary-sentinel.js';
 // Set max output tokens for Claude SDK (used by summarizer)
 process.env.CLAUDE_CODE_MAX_OUTPUT_TOKENS = '20000';
 // Increase max listeners for concurrent API calls
@@ -23,6 +23,27 @@ async function processBatch(items, processor, concurrency) {
 }
 function sessionIdForSummary(exchanges) {
     return exchanges.find(exchange => exchange.sessionId)?.sessionId;
+}
+/**
+ * Summarize one conversation once it has gone quiescent, writing the summary with
+ * its coverage header. Preserves a prior summary on failure (an error sentinel is
+ * written only for a first-time failure). Returns the summary, or null when the
+ * conversation is skipped (not yet quiescent) or summarization fails.
+ */
+async function summarizeIfQuiescent(summaryPath, archivePath, exchanges, label) {
+    if (!isQuiescent(exchanges, Date.now()))
+        return null;
+    try {
+        const summary = await summarizeConversation(exchanges, sessionIdForSummary(exchanges));
+        writeSummary(summaryPath, archivePath, exchanges, summary);
+        console.log(`  ✓ ${label}: ${summary.split(/\s+/).length} words`);
+        return summary;
+    }
+    catch (error) {
+        writeErrorSentinelIfNew(summaryPath, error);
+        console.log(`  ✗ ${label}: ${error}`);
+        return null;
+    }
 }
 export async function indexConversations(limitToProject, maxConversations, concurrency = 1, noSummaries = false) {
     console.log('Initializing database...');
@@ -89,27 +110,10 @@ export async function indexConversations(limitToProject, maxConversations, concu
             }
             // Batch summarize conversations in parallel (unless --no-summaries)
             if (!noSummaries) {
-                const needsSummary = toProcess.filter(c => shouldQueueForSummary(c.summaryPath));
-                if (needsSummary.length > 0) {
-                    console.log(`  Generating ${needsSummary.length} summaries (concurrency: ${concurrency})...`);
-                    await processBatch(needsSummary, async (conv) => {
-                        try {
-                            const summary = await summarizeConversation(conv.exchanges, sessionIdForSummary(conv.exchanges));
-                            fs.writeFileSync(conv.summaryPath, summary, 'utf-8');
-                            const wordCount = summary.split(/\s+/).length;
-                            console.log(`  ✓ ${conv.file}: ${wordCount} words`);
-                            return summary;
-                        }
-                        catch (error) {
-                            // Write an error sentinel so the failure is retryable on a later run (#96).
-                            try {
-                                fs.writeFileSync(conv.summaryPath, formatErrorSentinel(error), 'utf-8');
-                            }
-                            catch { }
-                            console.log(`  ✗ ${conv.file}: ${error}`);
-                            return null;
-                        }
-                    }, concurrency);
+                const toSummarize = toProcess.filter(c => needsSummary(c.summaryPath, c.archivePath));
+                if (toSummarize.length > 0) {
+                    console.log(`  Generating ${toSummarize.length} summaries (concurrency: ${concurrency})...`);
+                    await processBatch(toSummarize, conv => summarizeIfQuiescent(conv.summaryPath, conv.archivePath, conv.exchanges, conv.file), concurrency);
                 }
             }
             else {
@@ -173,21 +177,9 @@ export async function indexSession(sessionId, concurrency = 1, noSummaries = fal
                 if (exchanges.length > 0) {
                     // Generate summary (unless --no-summaries)
                     const summaryPath = archivePath.replace('.jsonl', '-summary.txt');
-                    if (!noSummaries && shouldQueueForSummary(summaryPath)) {
+                    if (!noSummaries && needsSummary(summaryPath, archivePath)) {
                         fs.mkdirSync(path.dirname(summaryPath), { recursive: true });
-                        try {
-                            const summary = await summarizeConversation(exchanges, sessionIdForSummary(exchanges));
-                            fs.writeFileSync(summaryPath, summary, 'utf-8');
-                            console.log(`Summary: ${summary.split(/\s+/).length} words`);
-                        }
-                        catch (error) {
-                            // Write an error sentinel so the failure is retryable on a later run (#96).
-                            try {
-                                fs.writeFileSync(summaryPath, formatErrorSentinel(error), 'utf-8');
-                            }
-                            catch { }
-                            console.log(`Summary failed: ${error instanceof Error ? error.message : String(error)}`);
-                        }
+                        await summarizeIfQuiescent(summaryPath, archivePath, exchanges, file);
                     }
                     // Index
                     for (const exchange of exchanges) {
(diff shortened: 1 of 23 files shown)


minyek:
# PR update — bound failed-summary retries (`eab6331`)

Follow-up to the "preserve a prior summary on failure" behaviour here. Preserving
was right, but the failed re-summary had **no backoff and no give-up**, so an
un-summarizable transcript re-invoked the LLM on *every* sync, forever. The
first-failure `#96` path had the mirror problem (backed off ~1h but never gave
up — and since sync is `SessionStart`-driven, the 1h floor rarely even skips a
sync).

**Fix:** one shared policy for both paths — back off until the retry floor
(`EPISODIC_MEMORY_SUMMARY_ERROR_RETRY_HOURS`, 1h), then give up after
`EPISODIC_MEMORY_SUMMARY_MAX_ATTEMPTS` (**new**, default 5), keeping any existing
summary. Failure state (`{attempts, lastAttempt}`) lives in the error sentinel and
in the coverage header's `resummary` field; success clears it. Also: never-summarized
conversations now sort ahead of re-summary retries before `summaryLimit`, so
failing retries can't starve fresh work (`#91`). Drops the mtime-based timing and
the duplicated logic.

Giving up first-failures too is deliberate: transient errors recover well within
5 attempts (≥5h); persistent ones never recover, so retrying forever only wastes
calls — and the exchanges stay indexed regardless.

**⚠️ Reviewer note — on-disk format change.** The `#96` sentinel's second line
goes from a bare ISO timestamp to a JSON failure-state line (to carry the attempt
count). Backward compatible: `parseErrorSentinel` reads the legacy v1.4.2 form, so
existing sentinels keep working and rewrite on their next failure.

Supersedes description point 5 (`writeErrorSentinelIfNew` → `recordSummaryFailure`).
Full suite: **258 pass.**

obra:
Triage note: holding this for a deliberate look after the current release, not merging it in this batch. It's a genuine lifecycle **feature** (no summary until a conversation is idle ~1h, refresh-on-growth, bounded retries) that also re-architects the summary-sentinel format, so it deserves its own review rather than riding along with the bug-fix batch. The spend/wedge problems it partly targeted are now addressed on main independently: bounded/auth-aware failure handling, an archive-only fallback (#122), and a hard SDK timeout (#160, EPISODIC_MEMORY_SUMMARY_TIMEOUT_MS). Left open on purpose; the quiescence behavior change (recent conversations get no summary until idle) is the main thing to weigh.

obra (changes requested):
First: this has been open since June 7 with no maintainer response, and that's on us, not you. What follows is a real review rather than a triage note.

**Verdict: land it in pieces — not re-implement.** An earlier internal triage bucketed this as "design still settling", and having actually read it, that's wrong for two of the three behaviours. The premise holds, the core is clean, and most of this is well-tested work that conflicts with almost nothing.

## Verified before reviewing

- **The bug is real and still unfixed on current `main`.** `shouldQueueForSummary` returns `false` for any real summary; nothing in the recent 13-PR batch or v1.6.0 records coverage or re-queues on growth.
- **The rebase is small.** `src/summary-sentinel.ts` — the heart of this PR — doesn't conflict at all. Only `indexer.ts`, `sync.ts`, `verify.ts`, mostly on import lines.
- **You rebuilt `dist` properly.** I ran `tsc` and the emitted `dist/*.js` and `*.d.ts` are byte-identical to what's committed. That's the thing most contributors to this repo get wrong, and worth saying out loud.
- CI never ran (this PR predates it). I ran the equivalent: `tsc --noEmit` clean, `npm test` **258/258** — exactly the number you claimed.

## The suggested split

| # | Piece | Files | State |
|---|---|---|---|
| 1 | Coverage header + growth-driven refresh + display strip | `summary-sentinel.ts` (most), `search.ts`, `verify.ts`, 3 call sites | Land after F1, F4, F5, F7 |
| 2 | Bounded retries / failure policy | `summary-sentinel.ts` (failure half), `sync.ts` error path | Land nearly as-is; F8 at rebase |
| 3 | Quiescence gate + fair queue ordering | `sync.ts`, `indexer.ts` | **This** is the piece still settling — F2, F3, F6 |

## Blocking

**F1 — `indexUnprocessed` re-summarizes from tail-only exchanges, then stamps full coverage.**
`indexer.ts:349` pushes only exchanges past the DB high-water mark, and `:369` hands that same array to `summarizeIfQuiescent` → `summarizeConversation` → `writeSummary`.

On `main` this is latent, because `shouldQueueForSummary` refuses to overwrite a real summary. This PR removes that guard for grown conversations, and `indexUnprocessed` *does* detect growth (`:337-339`). So a complete summary gets replaced by a summary of only the newest turns, with `buildCoverage` stamping `bytes` = full archive size — the header then asserts coverage the summary never had, and nothing else grew, so it never self-corrects.

That's arguably worse than the bug being fixed. Fix by re-parsing the full conversation in this path, or excluding it from re-summary.

**F2 — the quiescence skip happens *after* the `summaryLimit` slice.**
`sync.ts:208` slices to 10; `:229` then `continue`s on non-quiescent. So active conversations consume budget slots and produce nothing — and because they're never-summarized, `orderFreshBeforeRetry` sorts them into the *fresh* group at the head, stably, so the same active files pin the head every sync.

Reproduced with real `syncConversations` (summarizer mocked): 10 fresh + 1 idle → `summarized=0`, `llmCalls=0`, idle conversation never summarized. This re-creates the #91 head-of-queue failure through a new door.

Fix: count *produced* summaries rather than slicing candidates, or pre-filter on the **source** file's mtime. Your reasoning that the *archive* copy's mtime is unreliable is correct — the source's isn't.

**F3 — the headline behaviour has no test that can fail.**
I deleted both quiescence guards — `sync.ts:229` and `indexer.ts:55` — and the suite stayed **258/258 green**. Verified twice, independently.

The cause is that every sync-level test sets `EPISODIC_MEMORY_SUMMARY_QUIESCENCE_HOURS = '0'` to disable the gate. The pure `isQuiescent` predicate is tested, and tested well; the defer at every call site isn't.

## Non-blocking

**F4** — `search-summary-strip.test.ts` says the search loader must never show the `__COVERAGE__` line, but never imports `search.ts`. Reverting `search.ts:221` to `raw.trim()` leaves 258/258 green. The regression it guards is user-visible: a raw header line above every search result. (I checked the other `-summary.txt` consumers — `stats.ts`/`verify.ts` use the sentinel helpers, `show.ts` doesn't read it — so the strip itself is complete, just unguarded.)

**F5** — `ensureCoverageBaseline` stamps a legacy summary at *current* archive size, so an already-frozen partial summary is baselined as complete and never repaired. That's exactly the population your bug report describes. Users upgrading will expect their bad summaries to be fixed. A `baselined: true` flag `verify` can surface, or a one-shot `--resummarize-baselined`, would close it.

**F6** — `verify` doesn't know about quiescence: every conversation touched in the last hour now reports `No summary file`, and `repairIndex` (`verify.ts:163`) calls `summarizeConversation` unconditionally — so `verify --repair` freezes exactly the mid-session snapshot the defer exists to prevent. Bounded (growth refreshes it later), but self-inconsistent. Flip side: that bypass is currently the *only* escape from permanent give-up after `MAX_ATTEMPTS`, which answers "does a permanently-failing conversation go invisible?" — it doesn't, but only by accident. Worth making that intentional.

**F7** — `indexer.ts:138` copies to the archive only when it doesn't exist, so the archive never grows and `needsSummary` never re-queues. The refresh half is a no-op via the `index` CLI.

**F8 — a rebase trap.** `main` added a stop-batch path that writes `formatErrorSentinel` for *skips* (e.g. metered-API refusal), not failures. Your signature change forces that call site to be touched. Routing it naively through `recordSummaryFailure` would spend the 5-attempt budget on conversations never actually attempted. Probably `{attempts: 0, lastAttempt: now}`.

## What's good, and it's most of it

The retry and refresh halves are genuinely well tested — five mutations all caught red, including removing the give-up cap, the retry floor, and the growth gate. `sync-resummary.test.ts` drives the *real* `syncConversations` with only the LLM mocked and asserts byte-level coverage against a real `statSync`. That is the right shape, and it's precisely the shape F3 and F4 are missing — so you clearly know how to build it.

The legacy handling is careful in ways that are easy to skip: the ISO fallback in `parseErrorSentinel` so sentinels already on disk keep backing off; avoiding a mass re-summarization on upgrade; stripping a corrupt header rather than leaking it into the body. And the comments explain *why* — the note that the archive copy's mtime is reset every sync, so idleness must come from the last exchange's timestamp, is exactly the reasoning that's expensive to rediscover.

## One design question, worth deciding before piece 3

Piece 1 alone fixes the reported bug. But without the defer it costs one LLM call per sync per growing conversation, where today it costs one ever. So the defer isn't cosmetic — it's the cost control for the refresh.

The description rejects a minimum-growth-delta threshold on the grounds that the coverage scan is only ~23ms. That optimizes the wrong cost: the scan is free, the LLM call isn't. A delta threshold — re-summarize when the transcript grew by more than N bytes or 25% — is the direct lever on the cost the defer is paying for, and it has none of F2's, F3's or F6's problems. If you took that, piece 3 might shrink to just `orderFreshBeforeRetry`.

**Bottom line:** F1 needs fixing before anything lands. F2 and F3 are why the defer half should go last, on its own, with an integration test that doesn't set `QUIESCENCE_HOURS=0`. Everything else is landable. Sorry again about the three months.

PR #110, opened 2026-06-07
fix(summarizer): summarize conversations that end on an extended-thinking turn

minyek:
# Summarizer fails permanently on conversations whose last assistant turn used extended thinking

## Description

To summarize a short conversation (≤15 exchanges) the summarizer **resumes the session** via the Claude Agent SDK and asks for a `<summary>`. If that session's latest assistant message contains `thinking` / `redacted_thinking` blocks (extended thinking was on), resume replays that turn and the API rejects the continuation with a deterministic **400** — those blocks "cannot be modified". The conversation is written an error sentinel and retried, identically, on every sync forever — re-parsing a ~600 KB transcript and re-spawning a Claude subprocess each cycle.

Log fingerprint:

```
⚠️  Errors: 1
  <archive>/<uuid>.jsonl:
    Summary generation failed: Summarizer SDK error: success (session <uuid>)
```

`success` is misleading: the SDK reports this as an `SDKResultSuccess` with `subtype: 'success'` but `is_error: true`, `api_error_status: 400`, and the real message in `result`. The code discarded all of that and logged only the meaningless subtype.

Isolated but real: in an archive of 2,129 summaries this was the *only* errored conversation — the one that both ended on a thinking-block turn and was short enough (6 exchanges) to take the resume path. Size is not the cause; 500 MB+ transcripts summarize fine via the non-resume hierarchical path.

## Environment

- episodic-memory v1.4.2
- `@anthropic-ai/claude-agent-sdk` bundled with v1.4.2

## Root Cause

The Anthropic API constraint is the trigger; two code defects turn a recoverable condition into a permanent, opaque failure.

**1. `callClaude` throws away the real error.** It built `SummarizerSdkError` from only `subtype`, discarding `message.result` (`"API Error: 400 … thinking blocks … cannot be modified"`) and `api_error_status` (`400`). `subtype: 'success'` + `is_error: true` is a documented `SDKResultSuccess` combination — the agent loop finished, but the turn hit an API error.

**2. The non-resume fallback never fired.** `isResumeFailure` matched only `subtype === 'error_during_execution'`. This case is `'success'`, so no fallback ran and the error propagated to a sentinel. The non-resume path would summarize it fine — it never replays the thinking blocks.

Reproduced (resume the offending session, logging the raw result):

```json
{ "subtype": "success", "is_error": true, "api_error_status": 400, "num_turns": 1,
  "result": "API Error: 400 messages.1.content.25: `thinking` or `redacted_thinking` blocks in the latest assistant message cannot be modified. ..." }
```

## Fix Summary

Two parts; see the diff for code.

1. **Surface the real error.** `SummarizerSdkError` now carries `apiErrorStatus` and the API error text and renders them in the message, so the log shows the actual 400 instead of `success`.
2. **Broaden the non-resume fallback.** `isResumeFailure` now also treats an HTTP **400** on a resume call as a resume failure. Our summary prompt is plain, well-formed text, so a 400 there can only come from the replayed history; the non-resume path (fresh session, transcript as text) sidesteps the immutable thinking blocks. Non-resume statuses (401/403/429/5xx) still propagate without a futile retry.

Tests added: `test/summarizer-options.test.ts` (unit coverage for the broadened `isResumeFailure` and the diagnostic `SummarizerSdkError` fields/message) and `test/summarizer-resume-fallback.test.ts` (integration via a mocked SDK — the thinking-block 400 reproduced exactly, and recovery through the non-resume path).

## Fix Verified

Full suite passes (213/213). The integration test reproduces the exact SDK result shape from the real failure (`subtype: 'success'`, `is_error: true`, `api_error_status: 400`) and asserts the summary is recovered via the non-resume retry. Validated end-to-end against the original failing session (`fd2f1897`) on `claude-agent-sdk` 0.2.141: the resume reproduces the 400 as that exact result shape, `SummarizerSdkError` surfaces it, `isResumeFailure` routes it, and the non-resume fallback returns a correct summary. The diagnostic log now reads:

```
resume failed for <uuid> (Summarizer SDK error: success (HTTP 400) (session <uuid>): API Error: 400 … `thinking` … cannot be modified …); retrying without resume
```

Also in this PR: `test(show)` makes a timestamp assertion locale/timezone-independent — a pre-existing failure on a clean checkout under non-US locales (the assertion hardcoded a US/ISO date regex), unrelated to the summarizer but required to get the suite green.

### DIFF
--- a/dist/summarizer.d.ts
+++ b/dist/summarizer.d.ts
@@ -1,17 +1,26 @@
 import { ConversationExchange } from './types.js';
 /**
- * Thrown by callClaude when the SDK yields an `is_error: true` result message.
- * Carries the SDK's `subtype` and `session_id` as typed fields so callers can
- * dispatch on structural metadata rather than parsing error message text.
+ * Thrown by callClaude on an `is_error: true` result. Carries `subtype`,
+ * `session_id`, `api_error_status`, and the error text so callers can dispatch
+ * on structure and logs show the real failure, not a bare subtype. The SDK
+ * pairs `subtype: 'success'` with `is_error: true` when the loop finished but
+ * the turn hit an API error (carried in `result` / `api_error_status`).
  */
 export declare class SummarizerSdkError extends Error {
     readonly subtype: string;
     readonly sessionId?: string | undefined;
-    constructor(subtype: string, sessionId?: string | undefined);
+    readonly apiErrorStatus?: number | null | undefined;
+    readonly apiError?: string | undefined;
+    constructor(subtype: string, sessionId?: string | undefined, apiErrorStatus?: number | null | undefined, apiError?: string | undefined);
 }
 /**
- * True when the SDK's reported failure subtype indicates resume couldn't find
- * the session — the trigger for the non-resume fallback in summarizeConversation.
+ * True when the resume continuation itself failed — the trigger for the
+ * non-resume fallback in summarizeConversation. Two signals:
+ * - `error_during_execution`: the SDK couldn't resume (e.g. recorded cwd gone).
+ * - HTTP 400: the API rejected the replayed history, canonically `thinking`
+ *   blocks it forbids modifying on continuation. Our prompt is plain text, so a
+ *   400 on resume can only come from the replayed turns. Other statuses aren't
+ *   resume-specific and propagate.
  */
 export declare function isResumeFailure(error: unknown): boolean;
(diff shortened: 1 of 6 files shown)


obra:
Thanks @minyek! Merged to main as `f06748a`. Folded into a consolidated summarizer error-handling pass: the real SDK result text and status now surface instead of the contradictory "error: success".

Issue #42, opened 2025-12-30
Subagent conversations (Warmup) polluting search results despite is_sidechain support


He11oJack:
# Title

Subagent conversations (Warmup) polluting search results despite is_sidechain support

---

# Description

**Describe the bug**
Search results include subagent conversations (primarily "Warmup" initialization messages from Task tool) even though the `is_sidechain` column exists in the database schema and is correctly populated during parsing. The search queries in `search.js` do not filter on this column.

**To Reproduce**
1. Use Claude Code with episodic-memory plugin (v1.0.15)
2. Perform various tasks that spawn subagents via the Task tool
3. Run `episodic_memory_search` with any query (e.g., "authentication")
4. Observe that results include many `agent-*.jsonl` files with "Warmup" snippets

**Expected behavior**
Subagent conversations (`isSidechain: true`) should be excluded from search results by default, or a filter option should be available.

**Logs**
```
episodic_memory_search("project name")

Results:
1. [Project, 2025-12-30] - 24% match
   "Warmup"
   Lines 1-2 in .../agent-a045522.jsonl (1.6KB, 2 lines)

2. [Project, 2025-12-30] - 19% match
   "Warmup"
   Lines 1-2 in .../agent-aac9b21.jsonl (1.7KB, 2 lines)

... (many similar "Warmup" results)
```

**Additional context**
The infrastructure for filtering sidechains is already implemented but not used in search:

| Component | File | Status |
|-----------|------|--------|
| Schema column | db.js:59 | ✅ `is_sidechain BOOLEAN DEFAULT 0` |
| Index | db.js:102 | ✅ `CREATE INDEX idx_sidechain` |
| Parser extraction | parser.js:112 | ✅ `isSidechain: parsed.isSidechain` |
| Insert statement | db.js:124 | ✅ `exchange.isSidechain ? 1 : 0` |
| Search filter | search.js:49-51, 70-71 | ❌ **Missing** |

Proposed fix: Add `AND e.is_sidechain = 0` to WHERE clauses in `search.js`, or add an optional `include_sidechains` parameter (default: false).

Related: Issue #15 addressed this for the summarizer agent via exclusion marker, but general Task tool subagents are still indexed and returned in search results.

*Note: Analyzed and written with Claude Code.*

obra:
Fixed in commit fdb8b7d — added `AND e.is_sidechain = 0` to both vector and text search queries in src/search.ts. Sidechain/subagent conversations are now filtered out of search results.

[...]

PR #111, opened 2026-06-08
perf: skip copying and indexing sidechain conversations to reduce disk usage

minyek:
# perf: skip copying and indexing sidechain conversations to reduce disk usage

## Description

The primary goal of this PR is to **reduce episodic-memory's on-disk footprint**. The plugin copies and indexes sidechain conversations (subagent dispatches, Warmup/`ai-title` stubs, prompt-suggestion sidechains) that search already excludes and never returns. In my install that was **~40% of the SQLite DB and ~25% of the conversation archive** — disk spent on data no query can reach. This PR stops copying and indexing them so that storage isn't consumed going forward. It also avoids the embedding work for those exchanges, but the motivation is storage, not speed — sync latency is roughly unchanged. If a sidechain type later proves worth searching, it can be added with summarisation and query support.

- `sync.ts` copies every discovered `.jsonl` from `~/.claude/projects/` into `~/.config/superpowers/conversation-archive/`, including nested subagent dispatches at `<project>/<session>/subagents/agent-*.jsonl` and top-level harness stubs at `<project>/agent-*.jsonl` (Warmup, prompt-suggestion sidechains, `ai-title`).
- `indexer.ts` then embeds every parsed exchange into a 384-float vector and inserts rows into `exchanges` and `vec_exchanges`, regardless of `isSidechain`.
- `search.ts` unconditionally excludes them via `AND e.is_sidechain = 0` in both vector and text query paths.

This is a storage change, not a behavior change: search results are identical before and after, because the `AND e.is_sidechain = 0` filter (originally requested in #42) already hides these exchanges from every query and has no off switch. That left an asymmetry — the data is filtered out of every result, yet still embedded, written to the DB, and copied to disk. This PR removes that wasted storage by stopping at copy time and index time, matching the filter that search already applies.

## Why These Files Exist

All `isSidechain: true`, all legitimate Claude Code artifacts:

- **Explicit subagent dispatches** — Agent/Task tool transcripts under `<project>/<parent-session>/subagents/`. The agent's report-back already lives in the parent transcript and is searchable; the sidechain holds only the agent's internal trace (files read, intermediate reasoning).
- **Warmup and harness records** — top-level `<project>/agent-aXXX.jsonl`. Prompt-cache warmup stubs (`{"content":"Warmup"}`), prompt-suggestion sidechains, `ai-title` stubs. Most parse to 0 exchanges.
- **`/btw` asides** — Claude Code's `/btw` ("by the way") command asks a quick question off the main thread; its records are `isSidechain: true` and never join the main conversation. We considered surfacing these but intentionally don't: like every sidechain here they're already excluded from search by the `is_sidechain = 0` filter, so this PR doesn't change their searchability, and they're a small volume of deliberately ephemeral content — `/btw` exists precisely to keep asides out of history — so summarising and indexing them for search isn't worth the cost.

None have semantic value for cross-session search *as currently surfaced*.

## Change

Two layers:

**1. Skip sidechain exchanges at index time.** A single exported `indexExchanges()` in `src/indexer.ts` owns the embed + insert loop and short-circuits on `isSidechain`, skipping both the embedding generation and the DB insert — symmetric with the existing `AND e.is_sidechain = 0` filter in `src/search.ts`. All five indexing paths route through it (full reindex, single session, unprocessed backlog, **background sync**, and **verify/repair**), so the filter can't be applied to some paths and missed on others. The two that matter most are `syncConversations` — the background path that runs on every session start — and `repairIndex`; both would otherwise index inline sidechain exchanges from mixed-content files. The helper returns the inserted (non-sidechain) count, which the user-facing "Exchanges: N" logs report.

**2. Skip sidechain files at copy time.** New `isSidechainFile()` in `src/sync.ts` reads the file in chunks and inspects whole (newline-terminated) records until one carries an `isSidechain` flag, returning that value. It reads **complete records** rather than a fixed byte window, so a subagent file whose opening record is large (a big initial prompt with attachments) is classified correctly instead of truncated; the scan is bounded (1 MB) so a pathologically large file is never read in full. Conservative on every error path (unreadable file, an unparseable record, or no record carrying the field within the cap) so ambiguous and mixed-content files still get copied and handled by the per-exchange filter above. The check is gated behind a cheap mtime comparison, so unchanged files aren't re-read on every sync. Sidechain files are counted as `skipped`, never copied, and don't enter `filesToIndex` or `filesToSummarize`.

Schema groundwork for the `is_sidechain` column stays in place, so a future opt-in feature could lift the search filter and re-index from archive.

## Tests

`test/indexer-sidechain.test.ts` (new, 2 tests) — the manual reindex path (`indexUnprocessed`):

- Mixed transcript (one sidechain + one non-sidechain exchange) → `COUNT(*) FROM exchanges` is 1, `COUNT(*) WHERE is_sidechain = 1` is 0.
- All-sidechain transcript → `COUNT(*) FROM exchanges` is 0.

`test/sync.test.ts` (5 new tests):

- Top-level `agent-a01.jsonl` Warmup stub → not copied.
- Nested `<session>/subagents/agent-a01.jsonl` subagent dispatch → not copied.
- Regular non-sidechain conversation → copied as before.
- Mixed-content file (non-sidechain first record + inline sidechain exchanges) synced **with indexing enabled** → file is copied, but only the non-sidechain exchange lands in the DB. This pins the background `syncConversations` path, which the indexer test above doesn't exercise; it indexed both rows before this fix.
- Sidechain file whose first record is larger than the read chunk → still skipped at copy time (guards the complete-record read described above).

The sync test fixture now sets `EPISODIC_MEMORY_CONFIG_DIR` to isolate from any pre-existing `exclude.txt` that would have pre-filtered `subagents/` dirs at discovery and masked the test signal. Existing `test/incremental-indexing.test.ts` and `test/exclude-nested.test.ts` still pass.

## Validation

Beyond the unit suite, the change was exercised against real transcripts in an isolated scratch environment (real files read-only; scratch DB and archive, summaries off):

- A 41-file sample (32 subagent sidechains + 9 main conversations) synced to **9 copied, 32 skipped, 0 sidechain rows indexed**; the main conversations were indexed and returned from search as expected, and no real conversation was mis-skipped.
- This is where the large-first-record case surfaced. One subagent file had an 8 401-byte opening record. The earlier fixed-window peek truncated it mid-record, mis-classified it as non-sidechain, and copied it — search stayed correct (the per-exchange filter indexed 0 of its rows), but the file-level archive saving was lost. Reading complete records fixes it: the file is now skipped at copy time. That motivated layer 2's complete-record read and is locked in by a regression test.

## Environment

- episodic-memory v1.4.2

## For existing installs

This PR only prevents *new* sidechain data. Existing archived files and DB rows from prior versions stay until cleaned up. To reclaim space, run from the plugin directory:

```sh
# 1. Delete sidechain rows from the DB (handles vec_exchanges and tool_calls).
node -e "
  const { initDatabase, deleteExchange } = require('./dist/db.js');
  const db = initDatabase();
  const ids = db.prepare('SELECT id FROM exchanges WHERE is_sidechain = 1').all();
  const tx = db.transaction(rows => { for (const {id} of rows) deleteExchange(db, id); });
  tx(ids);
  db.exec('VACUUM');
  console.log('Deleted ' + ids.length + ' sidechain rows');
"

# 2. Remove archived nested-dispatch directories.
find ~/.config/superpowers/conversation-archive -type d -name subagents -exec rm -rf {} +
```

Step 2 covers `<session>/subagents/` dirs; top-level `<project>/agent-aXXX.jsonl` Warmup stubs from prior versions are smaller and can be left in place or removed manually. After this PR lands, no new sidechain files are written, so the cleanup is one-shot.

Safe to skip — search results are unaffected either way (sidechain rows weren't returned before, aren't now). Worth running if disk pressure matters; ~40% DB and ~25% archive reduction in my install.

### DIFF
--- a/dist/indexer.d.ts
+++ b/dist/indexer.d.ts
@@ -1,3 +1,16 @@
+import { initDatabase } from './db.js';
+import { ConversationExchange } from './types.js';
+/**
+ * Embed and insert exchanges, skipping sidechains (subagent dispatches, Warmup
+ * stubs, prompt-suggestion sidechains). Search excludes them via
+ * `AND is_sidechain = 0`, so indexing them would embed and store rows no query
+ * can ever return. Returns the number of exchanges actually inserted.
+ *
+ * This is the single place that decides what gets indexed — every indexing
+ * path (full reindex, single session, unprocessed backlog, background sync,
+ * verify/repair) routes through here so the sidechain filter can't be missed.
+ */
+export declare function indexExchanges(db: ReturnType<typeof initDatabase>, exchanges: ConversationExchange[]): Promise<number>;
 export declare function indexConversations(limitToProject?: string, maxConversations?: number, concurrency?: number, noSummaries?: boolean): Promise<void>;
 export declare function indexSession(sessionId: string, concurrency?: number, noSummaries?: boolean): Promise<void>;
 export declare function indexUnprocessed(concurrency?: number, noSummaries?: boolean): Promise<void>;
--- a/dist/indexer.js
+++ b/dist/indexer.js
@@ -24,6 +24,28 @@ async function processBatch(items, processor, concurrency) {
 function sessionIdForSummary(exchanges) {
     return exchanges.find(exchange => exchange.sessionId)?.sessionId;
 }
+/**
+ * Embed and insert exchanges, skipping sidechains (subagent dispatches, Warmup
+ * stubs, prompt-suggestion sidechains). Search excludes them via
+ * `AND is_sidechain = 0`, so indexing them would embed and store rows no query
+ * can ever return. Returns the number of exchanges actually inserted.
+ *
+ * This is the single place that decides what gets indexed — every indexing
+ * path (full reindex, single session, unprocessed backlog, background sync,
+ * verify/repair) routes through here so the sidechain filter can't be missed.
+ */
+export async function indexExchanges(db, exchanges) {
+    let inserted = 0;
+    for (const exchange of exchanges) {
+        if (exchange.isSidechain)
+            continue;
+        const toolNames = exchange.toolCalls?.map(tc => tc.toolName);
+        const embedding = await generateExchangeEmbedding(exchange.userMessage, exchange.assistantMessage, toolNames);
+        insertExchange(db, exchange, embedding, toolNames);
+        inserted++;
+    }
+    return inserted;
+}
 export async function indexConversations(limitToProject, maxConversations, concurrency = 1, noSummaries = false) {
     console.log('Initializing database...');
     const db = initDatabase();
@@ -117,12 +139,7 @@ export async function indexConversations(limitToProject, maxConversations, concu
             }
             // Now process embeddings and DB inserts (fast, sequential is fine)
             for (const conv of toProcess) {
-                for (const exchange of conv.exchanges) {
-                    const toolNames = exchange.toolCalls?.map(tc => tc.toolName);
-                    const embedding = await generateExchangeEmbedding(exchange.userMessage, exchange.assistantMessage, toolNames);
-                    insertExchange(db, exchange, embedding, toolNames);
-                }
-                totalExchanges += conv.exchanges.length;
+                totalExchanges += await indexExchanges(db, conv.exchanges);
                 conversationsProcessed++;
                 // Check if we hit the limit
                 if (maxConversations && conversationsProcessed >= maxConversations) {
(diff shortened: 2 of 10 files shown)


obra:
Closing in favor of #128, which the maintainer chose as the sidechain direction: subagent/Workflow conversations are now **searchable** (de-ranked below main-thread matches, opt out with include_sidechains=false) rather than dropped from the index. That landed on main (`489293b`). Thanks @minyek — the recall-starvation concern you raised is addressed by #128's over-fetch (k = limit*3), which also fixes the date-filter emptiness in #126.

PR #113, opened 2026-06-17
feat: Cursor support — live agent transcripts + legacy state.vscdb import

vicnaum:
Adds Cursor as a third supported harness alongside Claude Code and Codex — search, index, summarize, and `show` all work for Cursor conversations.

## What's included

**Live transcripts.** Cursor writes per-session agent transcripts to
`~/.cursor/projects/<slug>/agent-transcripts/<uuid>/<uuid>.jsonl` as
`{role, message:{content:[text|tool_use]}}` lines. Added harness detection
(role+message with no top-level `type`; `status`/`error` noise lines skipped so
they can't misdetect as Claude) and a parser: `<user_query>` unwrapping,
tool_use extraction, sessionId from the filename UUID. Transcripts carry no
timestamps or cwd, so timestamps fall back to file mtime (now preserved by
sync's `copyIfNewer`) and cwd is recovered from tool-call inputs with the path
slug as fallback.

**Legacy backfill.** Conversations older than the agent-transcripts era exist
only in Cursor's global `state.vscdb`. New `import-cursor-history` command opens
it read-only (better-sqlite3, already a dep), exports each conversation as
parser-compatible JSONL with real per-message timestamps, skips composers that
already have a live transcript, and attributes projects to the enclosing git
root. Output goes to a staging dir scanned as a conversation source; idempotent,
with `--force` to re-export.

**`show` rendering.** Added a Cursor branch to the markdown/HTML renderers
(they previously knew only Claude/Codex, so Cursor transcripts rendered empty).

## Also fixed along the way

- Summarizer attempted `claude --resume` on every non-Claude session; gated it
  on harness so Cursor/Codex sessions don't fail a resume before falling back.
- Sync re-queued summarizer-spawned message-less stub transcripts every run
  (feedback loop); message-less transcripts are now skipped.
- Subagent transcripts (`agent-<hex>.jsonl`, no UUID in the filename) were
  indexed but never summarized; they're now queued (summarized from text).

## Testing

229 tests pass (added Cursor parser, legacy-import, sync, and show coverage).
Validated on a real archive: 1,911 composers → 629 legacy exported, 270 deduped
against live transcripts, ~1,000 empty drafts skipped, 0 errors; full
Claude+Codex+Cursor index searchable end-to-end. `dist/` rebuilt per repo
convention (`npm ci && npm run build`).


Necmttn:
For the Cursor importer, I would keep provenance visible on every normalized conversation row.

Useful fields: source kind (`live_transcript|legacy_state_vscdb`), source path or sqlite row id, byte offset / message id, parser version, timestamp provenance (`message|mtime|fallback`), cwd provenance (`tool_input|slug|unknown`), dedupe key, and parse warnings. That makes later search results explainable when Cursor changes transcript shape again.

Generated with ax - https://github.com/Necmttn/ax

obra:
Thanks @vicnaum! Merged to main as `30916f8`. Cursor support (live transcripts + opt-in state.vscdb backfill) is in, following the Codex harness pattern. Rebuilt dist and re-ran the suite green.

PR #115, opened 2026-06-24
fix(search): over-fetch vector candidates so the sidechain filter can't starve results

iceinvein:
## Problem

Vector/semantic search returns far fewer results than `limit`, often zero at small limits, while text search works fine. On a ~31k-exchange index (~13% sidechain), `mode: "vector"` returned: limit 3 -> 0, limit 10 -> 2, limit 25 -> 11.

## Cause

In `searchConversations`, the vec0 KNN query is:

```sql
WHERE vec.embedding MATCH ? AND k = ?
  AND e.is_sidechain = 0
```

vec0 applies the `k` cutoff before the `WHERE`, but `k` is only inflated (`limit * 3`) when metadata filters are present, never for the always-on `is_sidechain = 0` filter. Subagent (sidechain) exchanges crowd the nearest neighbours and get removed after the cut, leaving fewer than `limit` results (often none at small limits). Text search is unaffected because it does not use `k`.

The `is_sidechain = 0` filter was added in #42; this is the latent recall regression introduced alongside it.

Related: #111 (skip indexing sidechains) would hide this for fresh installs, but its stated assumption that search results are identical before and after does not hold for the vector path, and it does not fix the query for any index that still contains sidechain rows.

## Fix

Over-fetch candidates with headroom for the always-on sidechain filter (`limit + 100`, or `limit * 3` when metadata filters are also active) and trim to `limit` after filtering (the existing trim handles the upper bound). This matches the existing `limit * 3` idiom; a fully guaranteed `limit` would require iterative `k` expansion, which could be a follow-up.

## Test

Adds `test/search-sidechain-recall.test.ts`: indexes one real exchange among 20 sidechain neighbours on the query topic, then asserts vector search still returns it at `limit: 3` and never returns sidechain rows. The test fails on the previous `k = limit` behaviour and passes with the fix.

## Verification

- Full suite: 208 passed, 1 pre-existing failure (`show.test.ts` date-format assertion) that also fails on clean `main`, unrelated to this change.
- `npm run build` clean.

### DIFF
--- a/dist/mcp-server.js
+++ b/dist/mcp-server.js
@@ -25111,7 +25111,7 @@ async function searchConversations(query, options = {}) {
   if (mode === "vector" || mode === "both") {
     await initEmbeddings();
     const queryEmbedding = await generateQueryEmbedding(query);
-    const k2 = hasMetadataFilters(options) ? limit * 3 : limit;
+    const k2 = Math.max(hasMetadataFilters(options) ? limit * 3 : limit, limit + 100);
     const stmt = db.prepare(`
       SELECT
         ${EXCHANGE_SELECT_COLUMNS},
--- a/dist/search.js
+++ b/dist/search.js
@@ -123,11 +123,17 @@ export async function searchConversations(query, options = {}) {
     const { sql: filterClause, params: filterParams } = buildSearchFilters(options);
     if (mode === 'vector' || mode === 'both') {
         // Vector similarity search.
-        // vec0 applies KNN before WHERE, so when extra metadata filters are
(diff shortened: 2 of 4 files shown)


obra:
Closing in favor of #128, which the maintainer chose as the sidechain direction: subagent/Workflow conversations are now **searchable** (de-ranked below main-thread matches, opt out with include_sidechains=false) rather than dropped from the index. That landed on main (`489293b`). Thanks @iceinvein — the recall-starvation concern you raised is addressed by #128's over-fetch (k = limit*3), which also fixes the date-filter emptiness in #126.

PR #117, opened 2026-06-30
feat: add opencode support

slandau3:
## Background

Episodic Memory already worked well for Claude Code and Codex because both expose transcript-shaped conversation history. opencode stores conversations in SQLite, so we needed an adapter layer to make it fit the same archive/index/search pipeline.

## What Changed

- Added opencode session export from `~/.local/share/opencode/opencode.db` into generated JSONL transcripts
- Extended the parser/display stack to understand the opencode transcript format
- Added an opencode plugin entrypoint that runs background sync on `SessionStart`
- Added `episodic-memory doctor opencode` for setup diagnosis
- Added opencode-specific docs at `docs/OPENCODE.md`
- Kept the same MCP search/read surface so opencode can reuse the existing memory tools
- Updated the README to add opencode alongside Claude Code and Codex, mirroring the existing structure

## Key Files

- `src/opencode-transcripts.ts` — exports opencode sessions into generated JSONL
- `src/parser.ts` — parses opencode transcript records into exchanges
- `cli/opencode-plugin.js` — opencode plugin hook that launches background sync
- `src/sync.ts` — adds opencode transcript source and `opencode-doctor` checks
- `src/paths.ts` / `src/opencode-doctor.ts` — opencode environment checks
- `docs/OPENCODE.md` — install and configuration guide

## Decisions

- Chose a DB-to-JSONL export instead of indexing the SQLite DB directly, so opencode can use the same archive/search path as the other harnesses
- Kept the package in the existing repo rather than splitting it out, because the plugin reuses the same shared memory engine and MCP server
- Exposed the plugin as `./server` via package exports so opencode can load it cleanly
- Used the existing file-lock based sync guard so repeated `SessionStart` events do not spawn concurrent indexing runs

## Verification

- `npm run build` passes
- `npm test` passes (220 tests across 39 files)
- Live smoke test against a temp opencode SQLite DB, confirming export, copy, and index


bernardofortes:
+1 — real-world use case here: running Claude Code + Codex + OpenCode on the same machine with episodic-memory as the unified session-search layer. Claude Code and Codex sides are working great; OpenCode is the missing third. This PR is exactly the missing piece — happy to test the branch on a live setup with a ~11GB opencode.db if that helps review.

obra:
Thanks @slandau3! Merged to main as `30916f8`. opencode support is in, with the summarization path fixed to route via transcript text (and hasConversationContent taught to recognize opencode_message lines) so opencode conversations actually get summarized. Added a routing test.

Issue #97, opened 2026-05-19
No single-instance lock — concurrent SessionStart hooks pile up syncs on Windows (multi-worktree)


netoatr-glitch:
## Summary

`dist/sync-cli.js` has no concurrency control. The `SessionStart` hook fires `episodic-memory sync --background` on every new Claude Code session. On a setup with many simultaneous sessions (multiple git worktrees, terminals, sub-agents), N independent SessionStart events spawn N concurrent background sync workers. Each walks the same archive and tries to summarize the same backlog, and each summarization spawns a `claude` SDK subprocess.

On Windows this burst trips desktop-heap / session-object exhaustion; the spawned `claude` subprocesses fail DLL initialization and exit with **STATUS_DLL_INIT_FAILED (0xC0000142 / decimal 3221225794)**. Combined with #96 (no sentinel on error), it becomes a permanent error loop that re-spawns the workload on every session start.

## Evidence

```
2026-05-19T20:35:44.306Z [info] Starting background sync from pid 19880
2026-05-19T20:35:44.335Z [info] Starting background sync from pid 19460
2026-05-19T20:35:44.415Z [info] Starting background sync from pid 19044
```

Three workers started within 110ms. Process list at the time: 10 concurrent `dist/sync-cli.js` node procs, 18 `claude.exe` children. Errors:

```
Summary generation failed: Claude Code process exited with code 3221225794
Summary generation failed: Cannot read properties of undefined (reading 'match')
```

The reentrancy guard `shouldSkipReentrantSync()` (#87) correctly prevents a summarizer-spawned Claude from re-triggering sync. But it does not cover independent SessionStart events from different sessions — each has a clean `EPISODIC_MEMORY_SUMMARIZER_GUARD` env, so each spawns a full sync.

## Suggested fix

Single-instance lockfile in `sync-cli.js`. Acquire before `syncAll()`, release on exit. Detect stale locks by (a) dead PID via `process.kill(pid, 0)` and (b) mtime older than ~30 min. Fail-open on unexpected FS errors so a weird state can never wedge sync permanently.

I implemented this locally and validated three scenarios:
- Lock held by alive process → second sync exits with `another sync (pid X) is running — skipping` (no work done, holder lock untouched)
- Stale lock (dead PID) → second sync clears it and takes over, releases on exit
- Normal acquire/release → no leftover lock file

Sketch (writes lock at `dirname(getSyncLogPath())/episodic-memory-sync.lock`, full version is ~50 LoC):

```js
[...]

PR #124, opened 2026-07-12
Cap onnxruntime intra-op threads on Apple Silicon to stop the embedding CPU peg

scottyallen:
## Problem

Embedding is done one exchange at a time (batch=1) throughout indexing, sync, and the re-embed migration. At that size `onnxruntime-node`'s default intra-op thread pool fans each tiny int8 matmul across every core. On Apple Silicon this pegs ~5 cores for almost no throughput gain, and it makes a bulk re-embed (e.g. the model migration) hog the whole machine. It's especially visible when several Claude Code sessions/worktrees trigger syncs, or when the machine is otherwise busy and the pool oversubscribes and fights itself.

## Fix

Cap the embedding session's thread pools when we know parallelism won't help:

- `intraOpNumThreads: 2` and `interOpNumThreads: 1`, applied only on `darwin && arm64`.
- Gated as a strict no-op elsewhere: when no cap resolves, the `pipeline()` call passes **no** `session_options` at all, so Linux/Intel/CI are byte-for-byte unchanged.
- Env override `EPISODIC_MEMORY_EMBED_THREADS`: a positive integer sets the cap; `0` / negative / non-numeric restores onnxruntime's default (off-switch). Parsed defensively.

Same model, same `q8` CPU execution provider, so **embeddings are bit-identical to the uncapped output — no re-index is triggered.**

## Measurements (M2 Max, `bge-small-en-v1.5` q8, batch=1, N=200)

| Config | per-item wall | cores (cpu/wall) | vs stored q8 vectors | re-index |
|---|---|---|---|---|
| Uncapped default (current) | 11.3 ms | **4.67** | baseline | - |
| **Capped intra=2 / inter=1** | 10.6 ms | **1.78** | cosine 1.000000 (exact) | **no** |

Same-or-faster wall time while using ~2.6x fewer cores. The gap widens under system load, because capping stops the pool from oversubscribing.

## Alternatives evaluated and rejected

- **CoreML EP** (`device: 'coreml'`) on the q8 model: ~9x *slower* (~105 ms/item), never actually leaves the CPU (still ~5 cores — the int8 graph partitions poorly), and produces different vectors (would force a re-index). Hard no.
- **WebGPU fp32** (dawn): genuinely offloads (~0.78 cores) but only at fp32, whose vectors differ from the stored q8 ones (full re-index) for no wall-clock gain, and it's experimental.

The thread cap is the only option that reduces CPU with zero behavioral change to the embeddings.

## Tests

- New unit test for `resolveIntraOpThreads()` covering positive override, `0`, non-numeric, and the platform default.
- Full suite green (211/211).

### DIFF
--- a/dist/embeddings.d.ts
+++ b/dist/embeddings.d.ts
@@ -1,4 +1,28 @@
 export declare const BGE_QUERY_PREFIX = "Represent this sentence for searching relevant passages: ";
+/**
+ * Resolve an intra-op thread cap for the embedding session, or null to leave
+ * onnxruntime at its default (one worker per core).
+ *
+ * Embedding is done one exchange at a time (batch=1) throughout indexing, sync,
+ * and the re-embed migration. At that size onnxruntime-node's default intra-op
+ * thread pool fans each tiny int8 matmul across every core, which on Apple
+ * Silicon pegs ~5 cores for almost no throughput gain and makes a bulk re-embed
+ * hog the whole machine. Capping intra-op threads keeps embedding off the
+ * user's cores; measured on an M-series Mac, a cap of 2 is as fast or faster
+ * than the uncapped default while using ~1.7 cores instead of ~5 (and pulls
+ * further ahead when the machine is otherwise busy, since it stops the pool
+ * from oversubscribing).
+ *
+ * This stays on the same CPU / q8 execution provider, so embeddings are
+ * bit-identical to the uncapped output — no re-index is triggered. (CoreML and
+ * WebGPU were evaluated and rejected: on this quantized model they either run
+ * far slower and never leave the CPU, or require an fp32 model whose vectors
+ * differ from the stored q8 ones and would force a full re-index.)
+ *
+ * Override with EPISODIC_MEMORY_EMBED_THREADS: a positive integer sets the cap;
+ * 0 (or any non-positive/invalid value) restores onnxruntime's default.
+ */
+export declare function resolveIntraOpThreads(): number | null;
 export declare function initEmbeddings(): Promise<void>;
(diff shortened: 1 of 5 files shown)


dkindlund:
Independent corroboration on an M3 Pro (5P + 6E), if useful — your 4.67-core figure matches almost exactly on different hardware.

Measured during a full index pass over a 4.5 GB / ~29,000-exchange corpus:

```
real 899s   user 4108s   →  4.57x parallelism
```

and per-process `%CPU` during sync ranged **300–476%**. So ~4.6 cores per embedding process, versus the 4.67 you measured on the M2 Max — same behavior, two different chips.

Before I upgraded, three concurrent SessionStart syncs on v1.2.0 put the machine at **~874% CPU (~9 cores)**.

One argument in favor of this PR that may not be obvious: the single-instance lock from #97 (shipped in v1.4.2) removes the *N×* multiplier, but not the *1×* cost. Post-lock I still see one process holding ~4.6 cores for the duration of a pass. This PR is what takes that remaining 4.6 → ~1.8, and the two fixes compose rather than overlap.

Batch=1 also means the wide thread pool is buying very little: at that size the per-op fan-out is mostly coordination overhead, which is consistent with your "bit-identical embeddings" result.

(Context on the workload that made this visible, including a separate non-incremental re-embed issue in the sync path: #152.)

obra:
Thanks @scottyallen! Merged to main as `dd5815f` (squashed as part of a 13-PR batch, with dist rebuilt on top). This caps onnxruntime intra-op threads on Apple Silicon to stop the embedding CPU peg, with an env override and a no-op on every other platform.

PR #128, opened 2026-07-16
fix: make subagent and workflow (sidechain) conversations searchable

d-walp:
## Problem

Both search paths append `AND e.is_sidechain = 0` (`src/search.ts`, vector + text), so every exchange recorded inside a subagent or `Workflow` run is excluded before any matching happens, and `SearchOptions` exposes no way to include them.

For orchestration-heavy workflows (parallel review panels, workflow-driven implementation) sidechains hold the *majority* of the substance — in such indexes the excluded rows can outnumber the searchable main thread severalfold. Semantic search returns the thin main-thread shell and silently misses the work — the worst failure mode for a memory tool, since it still returns *something*.

## Change

Sidechains are now **searched by default and de-ranked** just below equally-relevant main-thread matches, so the main thread still surfaces first but the substance is reachable:

- **vector:** `ORDER BY (vec.distance + is_sidechain * 0.05)` — a main-thread row wins ties/near-ties; a clearly-closer sidechain still ranks. Candidate pool widened to `limit * 3` since vec0 runs KNN before the WHERE/ordering.
- **text:** `ORDER BY is_sidechain ASC, timestamp DESC`.

Opt back into main-thread-only search with `include_sidechains: false` (MCP tool param) or `--exclude-sidechains` (CLI).

## Text-match contract (documented, behavior unchanged)

The `text` path's existing behavior is now documented and test-pinned: the whole query is matched as **one contiguous, case-insensitive substring** (`LIKE %query%`), with no tokenization — so `text` mode suits exact strings (SHAs, error codes) and `vector`/`both` suit natural-language queries. A future FTS5 / term-AND upgrade would improve `text` recall but is out of scope.

## Tests / build

`test/search-sidechain.test.ts` (9 tests, real indexer + embeddings + sqlite-vec): sidechain-only work found by default (vector + text); `include_sidechains: false` restores exclusion; equally-relevant main-thread rows outrank sidechains (both paths); default returns both; text verbatim-substring contract. Full suite 216/216. `dist/` rebuilt (this also refreshes `dist/version.js` from package.json 1.4.2, which was committed stale at 1.4.1).


HulkInTherapy:
Some field data for the excluded set, in case it is useful for the ranking
question here.

I forked episodic-memory's search into a local-first archive tool and ran it
over one developer machine's full Claude Code history. Of 227 transcripts
indexed, **197 were sidechains against 30 live sessions** — better than six
files of delegated work for every session the person remembered having. With
`AND e.is_sidechain = 0` in place, that is the majority of the corpus
unreachable, and it is the half that contains the actual implementation work:
the main thread is frequently a short "here is the plan, go" and a summary at
the end. Your de-ranking approach handles the case the flag I had prepared does
not, which is that the main thread should still win a tie without the substance
becoming invisible.

The fork is [potsherd](https://github.com/HulkInTherapy/potsherd) (MIT, credited
in NOTICE and in a port log). I had written up the same hard-coded filter as a
patch before finding this PR; that patch is now a document rather than a
submission, since this does it better. Nothing needed from you — just adding the
number.

obra:
Thanks @d-walp! Merged to main as `489293b`. Sidechain (subagent/Workflow) conversations are now searchable by default, de-ranked below main-thread matches, with include_sidechains=false to opt out. Synthesized with the merged multi-word text search (#144); dropped the stray fast-uri dist hunk on a clean rebuild.

PR #130, opened 2026-07-19
fix(show): pin conversation timestamps to en-US/UTC

### LINKED ISSUE #120: show.test.ts date assertion fails on non-US locales (ko-KR + KST)
`show command - markdown formatting > should format a simple user-assistant exchange` asserts `/9\/19\/2025|2025-09-19/`, but the rendered timestamp is locale/timezone dependent. On a ko-KR + Asia/Seoul machine it renders as `2025. 9. 20. 오전 2:34:29` (Korean locale format **and** UTC→KST day shift), so the suite can never fully pass here — 206/207 otherwise green on v1.4.2, macOS 15, node 26.

Suggested fix: pin the formatter, e.g. `toLocaleString('en-US', { timeZone: 'UTC' })` in show.ts (or inject a formatter and assert against it), so `show` output and the test agree regardless of the host locale.

arimu1:
## Summary
- Pin `toLocaleString('en-US', { timeZone: 'UTC' })` for conversation timestamps in `show.ts` (markdown + HTML + Codex paths).
- Tighten `show.test.ts` to assert the stable `9/19/2025` form.

Fixes #120

## Test plan
- [x] Assertion matches pinned UTC formatting for the fixture timestamp
- [ ] `npx vitest run test/show.test.ts` on a non-US locale if available

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/show.ts
+++ b/src/show.ts
@@ -84,7 +84,7 @@ export function formatConversationAsMarkdown(jsonl: string, startLine?: number, 
 
   for (let i = 0; i < messages.length; i++) {
     const msg = messages[i];
-    const timestamp = new Date(msg.timestamp).toLocaleString();
+    const timestamp = new Date(msg.timestamp).toLocaleString('en-US', { timeZone: 'UTC' });
     const messageId = msg.uuid || `msg-${i}`;
 
     // Skip user messages that are just tool results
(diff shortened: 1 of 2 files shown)


obra:
Thanks @arimu1! Merged to main as `30916f8`. Conversation timestamps are pinned to en-US/UTC in every renderer (markdown, HTML, Codex, Cursor, opencode), so display no longer day-shifts on non-US locales.

PR #133, opened 2026-07-22
fix: update hasMetadataFilters() to return true for date filters (#126)

### LINKED ISSUE #126: Date-filtered search returns empty: vec0 KNN runs before WHERE, and hasMetadataFilters() ignores after/before
**Version:** 1.4.2 (code unchanged on current `main`). Affects both the CLI and the MCP `search` tool — both go through `searchConversations()`.

## Symptom

`search` with `after` (or `before`) reliably returns "No results found", even when the index provably contains many matching rows:

- `search "<query>" --after <3 days ago>` → 0 results
- the same query without `--after` → returns (older) results fine
- `SELECT COUNT(*) FROM exchanges WHERE timestamp >= '<3 days ago>'` against `~/.config/superpowers/conversation-index/db.sqlite` → hundreds of rows

Observed on an index spanning ~3 months.

## Mechanism

The vector path in `searchConversations()` (src/search.ts:157-168) runs:

```sql
FROM vec_exchanges AS vec
JOIN exchanges AS e ON vec.id = e.id
WHERE vec.embedding MATCH ?
  AND k = ?
  AND e.is_sidechain = 0
  AND e.timestamp >= ?   -- from buildSearchFilters()
```

sqlite-vec's vec0 virtual table performs the KNN scan *first* and applies the remaining WHERE clauses to those k candidates afterwards. The code knows this — the comment at src/search.ts:151-152 says so, and k is widened when filters are present:

```ts
const k = hasMetadataFilters(options) ? limit * 3 : limit;
```

But `hasMetadataFilters()` (src/search.ts:53-55) only checks `project`/`session_id`/`git_branch`:

```ts
function hasMetadataFilters(options: SearchOptions): boolean {
  return Boolean(options.project || options.session_id || options.git_branch);
}
```

`after`/`before` are applied as WHERE filters (src/search.ts:27-34) but don't widen k. So a date-filtered search fetches the global top-`limit` (default 10) nearest neighbors from the *entire* index, then the date filter is applied to those 10. On a months-old index the odds that the global top-10 for a query all fall inside a narrow recent window are tiny → stable empty result.

Two layers to it:

1. `after`/`before` aren't counted as filters at all, so k isn't widened — the main bug.
2. Even when k *is* widened, `limit * 3` is too small for selective filters: on my index `--after <3 days ago>` passes 1,270 of 22,366 rows (~5.7%), so surfacing 10 post-filter matches needs k ≈ 175 by expectation, not 30.

Open PR #115 hits the same vec0 KNN-before-WHERE starvation from a different angle (the always-on `is_sidechain = 0` filter) and raises k to `max(..., limit + 100)`. That would soften this bug but not fix it: `after`/`before` would still not count as filters, and ~110 global candidates still starve on more selective date windows (e.g. a one-day window on a months-long index). The two fixes compose cleanly.

## Repro

On any index with a few months of history:

1. `search "<query matching recent work>" --after <3 days ago>` → "No results found"
2. Same query without `--after` → results (all older)
3. `sqlite3 ~/.config/superpowers/conversation-index/db.sqlite "SELECT COUNT(*) FROM exchanges WHERE timestamp >= '<3 days ago>'"` → hundreds of rows

## Suggested fix

Part 1 — count `after`/`before` as filters (src/search.ts:53-55):

```diff
 function hasMetadataFilters(options: SearchOptions): boolean {
-  return Boolean(options.project || options.session_id || options.git_branch);
+  return Boolean(
+    options.project || options.session_id || options.git_branch || options.after || options.before
+  );
 }
```

Part 2 — derive k from filter selectivity instead of a constant. The WHERE runs over the k KNN candidates, so (assuming the filter is roughly independent of semantic rank) the expected number of survivors is k·p, where p is the fraction of rows the filters pass; surfacing `limit` results needs k ≈ limit/p. Measured on my index (22,366 non-sidechain rows; a 3-day window passes 1,270 → p ≈ 5.7%), with limit = 10:

| k | expected survivors (k·p) |
|---|---|
| 10 — current: date filters don't widen k | ~0.6 → usually empty |
| 30 — `limit * 3` | ~1.7 |
| ~175 — `limit / p` | ~10 = limit |

p is one cheap indexed COUNT away, so k can be derived instead of guessed:

```ts
// sketch
const matching = countMatchingRows(db, options);  // COUNT(*) FROM exchanges WHERE <same filters>
if (matching === 0) return [];                    // filters exclude everything — skip vector search
const p = matching / totalRows;
const k = Math.min(totalRows, Math.ceil((limit / p) * 2));  // 2× margin over expectation
```

One trap discovered when implementing this: **sqlite-vec hard-caps KNN k at 4096** (`k value in knn query too large, provided 4613 and the limit is 4096`). A same-day `--after` window on my index derives k = 4613 (97 of 22,366 rows pass, p ≈ 0.4%) and the KNN query errors outright. But needing k > 4096 also means the filters are very selective by construction (p < 2·limit/4096), so the natural fallback is to skip KNN and compute exact distances over just the filtered rows:

```sql
SELECT ..., vec_distance_L2(vec.embedding, :query) AS distance
FROM exchanges AS e
CROSS JOIN vec_exchanges AS vec
WHERE vec.id = e.id
  AND e.is_sidechain = 0
  AND <same filters>
ORDER BY distance ASC
LIMIT :limit
```

Exact, immune to starvation by construction, and cheap at that selectivity (~0.2s for a 97-row subset of a 22k index). Two implementation notes: `CROSS JOIN` pins the join order so SQLite filters `exchanges` first and does point lookups into vec0 — with a plain `JOIN` the planner scanned the whole vec table (~5× slower here); and `vec_distance_L2` is the same metric vec0 KNN returns, so the distance→similarity conversion downstream is unchanged.

Implemented and verified locally on v1.4.2 (derived k + the exact-scan fallback): a 3-day `--after` window that returned 0 results now returns a full 10, and a same-day window that would error under the KNN cap returns 10 via the exact scan. The over-fetch is trimmed back to `limit` after ranking (src/search.ts:175-177), so response size is unchanged.

Happy to send a PR.

## Related

Filed separately as #127: the text branch matches the query as one literal substring (`LIKE '%query%'`), so in the default `both` mode it can't rescue vector starvation — any multi-word paraphrase returns nothing there either. That's why this issue presents as empty results even in `both`.

arimu1:
## Summary
Fixes issue #126 where `hasMetadataFilters()` in `src/search.ts` did not return `true` when `filter.after` or `filter.before` date options were specified.

When date filters are used without project/session/branch filters, vector search previously set `k = limit` instead of expanding to `k = limit * 3`. Because `sqlite-vec` applies KNN before the `WHERE` timestamp filter clause, candidates outside the top `limit` were excluded before date filtering was applied, causing missing or empty search results.

## Changes
- Updated `hasMetadataFilters()` in `src/search.ts` to check `options.after` and `options.before`.
- Added test coverage in `test/search-metadata-filters.test.ts` for vector search with date range filters.

## Verification
- Ran `npx vitest run test/search-metadata-filters.test.ts` (all 8 tests passed).

## AI Disclosure
This change was implemented with assistance from AI (Claude Code / Antigravity Agent).

Closes #126

### DIFF
--- a/src/search.ts
+++ b/src/search.ts
@@ -51,7 +51,7 @@ function buildSearchFilters(options: SearchOptions): { sql: string; params: unkn
 }
 
 function hasMetadataFilters(options: SearchOptions): boolean {
-  return Boolean(options.project || options.session_id || options.git_branch);
+  return Boolean(options.project || options.session_id || options.git_branch || options.after || options.before);
 }
 
 const EXCHANGE_SELECT_COLUMNS = `
--- a/test/search-metadata-filters.test.ts
+++ b/test/search-metadata-filters.test.ts
@@ -176,4 +176,18 @@ describe('search metadata filters', () => {
     expect(projects.has('project-a')).toBe(true);
     expect(projects.has('project-b')).toBe(true);
   });
+
+  it('filters by after and before date filters in vector search mode (#126)', async () => {
+    const resultsAfter = await searchConversations('authentication', { after: '2026-01-02', mode: 'vector', limit: 10 });
+    expect(resultsAfter.length).toBeGreaterThan(0);
+    for (const r of resultsAfter) {
+      expect(new Date(r.exchange.timestamp) >= new Date('2026-01-02T00:00:00.000Z')).toBe(true);
+    }
+
+    const resultsBefore = await searchConversations('authentication', { before: '2026-01-02', mode: 'vector', limit: 10 });
+    expect(resultsBefore.length).toBeGreaterThan(0);
+    for (const r of resultsBefore) {
+      expect(new Date(r.exchange.timestamp) <= new Date('2026-01-02T23:59:59.999Z')).toBe(true);
+    }
+  });
 });


anastasiiaanfimova:
Thanks for picking this up — the `hasMetadataFilters()` change is right.

Worth flagging what it leaves open: `k = limit * 3` is still a constant, and that constant is where the starvation actually comes from. Past a filter passing a fraction `p` of the table, surfacing `limit` rows needs `k ≈ limit/p`. A three-day window over a three-month index is `p ≈ 3%`, so `limit*3` yields roughly one row of the ten requested — the empty result becomes a sparse one rather than a correct one. Above vec0's 4096 cap it fails differently again.

I have that second half implemented and tested on v1.4.2 — derived `k` plus the exact-scan fallback, both sketched at the end of #126:

https://github.com/anastasiiaanfimova/episodic-memory/tree/fix/date-filter-knn-selectivity

`test/search-metadata-filters.test.ts` goes 8 → 11 cases, full suite 211/211.

Happy either way: fold the commit into this PR if you'd rather ship one change, or I open it as a follow-up on top of yours. Your call.

obra:
Thanks @arimu1! The date-filter emptiness (#126) is fixed on main via #128, which removed `hasMetadataFilters` and made vector search always over-fetch (k = limit*3) while applying after/before in the WHERE — so your PR's code change is subsumed. I kept your intent as a timezone-robust regression test guarding date-filtered vector search (`489293b`), crediting you as co-author. Closing as covered.

Issue #94, opened 2026-05-18
SessionStart sync failures are invisible: `npm rebuild ... 2>/dev/null || true` + `--background` hook hide native-binding errors from users


orionsky081:
## Summary

When `better-sqlite3` (or any native dep) fails to build on a user's machine, the failure is completely invisible from Claude Code's perspective. The user sees a vague "SessionStart hook error" with no indication that episodic-memory is involved, let alone which native module is broken. The plugin then silently stops archiving conversations for every future session until someone manually inspects `~/.config/superpowers/logs/episodic-memory.log`.

I lost ~30 minutes to this today before tracing it to the actual error. Diagnostics paste at the bottom.

This issue is about the **observability failure mode**, not about fixing my specific install (`npm rebuild better-sqlite3` resolved it locally). Prior install-side fixes (#56, #76/#86) won't help the next user who hits a different native-build failure — the silent-fail path will swallow that too.

## Environment

- Plugin version: 1.3.1
- Plugin source: `superpowers-marketplace`
- OS: Linux 6.6.87.2-microsoft-standard-WSL2 (Ubuntu under WSL2)
- Node: v22.22.2
- Claude Code: latest as of 2026-05-19

## What the user sees vs what's actually happening

**User-visible:** Claude Code surfaces `SessionStart:startup hook error`. No plugin name, no stack, no path to investigate. The session otherwise appears to start normally.

**Actually happening:**

1. `hooks/hooks.json` runs `node .../cli/episodic-memory.js sync --background`.
2. Parent exits 0 immediately (because `--background` detaches the child). From Claude Code's view, the hook succeeded.
3. The detached child tries `require('better-sqlite3')`, throws `Error: Could not locate the bindings file`, writes the stack to `~/.config/superpowers/logs/episodic-memory.log`, and dies.
4. Every subsequent SessionStart repeats step 1–3. Conversations stop being archived. The user has no idea.

The "SessionStart:startup hook error" Claude Code does show must be coming from somewhere else on the path (possibly a sibling hook or the non-JSON stdout `Sync started in background. Log: ...` line). Either way, the message points at no useful target.

## Root cause of the invisibility

Two design choices compound:

1. **`postinstall` swallows build errors.** The install path runs `npm rebuild better-sqlite3 2>/dev/null || true` — the `2>/dev/null || true` makes a failed rebuild indistinguishable from success. #41 alluded to this script's existence but the resolution there was about the macOS retry loop, not the silent-fail semantics.

[...]

PR #137, opened 2026-07-23
fix: degrade gracefully when the embedding backend can't load (#135)

### LINKED ISSUE #135: sharp/libvips dlopen fails (host without system libvips): @img hoisting split crashes SessionStart sync
## Summary

On a host **without** a system-wide libvips, the bundled `sharp@0.34.5` (pulled in transitively by `@huggingface/transformers`) fails to `dlopen` its bundled libvips, which **crashes the `SessionStart` sync hook** and any `episodic-memory.js sync` run.

Root cause is an npm hoisting split: `@img/sharp-linux-x64` (the native `.node`) gets hoisted to the plugin's top-level `node_modules/@img/`, but its bundled libvips `@img/sharp-libvips-linux-x64` stays **nested** under `node_modules/sharp/node_modules/@img/`. None of the prebuilt binary's `RPATH` entries bridge that gap, so `libvips-cpp.so.8.17.3` can't be found.

This is **distinct from #102** (that one is host *with* global libvips → sharp postinstall picks a *source* build). Here there is *no* system libvips and the *prebuilt* binary is the one that fails, purely due to the `@img` package placement. It's related to #94 in that the failure surfaces through the `SessionStart` `sync --background` hook.

## Environment

- OS: Manjaro Linux, x86-64
- Node: v26.4.0
- Plugin: `episodic-memory@1.4.2` (installed via `superpowers-marketplace`)
- No system libvips installed (`pacman -Q libvips` → not found; `ldconfig -p | grep vips` → empty)
- `sharp@0.34.5` via `@huggingface/transformers@4.2.0` (confirmed with `npm ls sharp`)

## Symptom

Claude Code reports at startup:

```
SessionStart:startup hook error
Failed with non-blocking status code: .../episodic-memory/1.4.2/node_modules/sharp/lib/sharp.js:120
```

Running the underlying command directly reproduces it and shows the sync **crashes with exit 1**:

```
$ node cli/episodic-memory.js sync
.../node_modules/sharp/lib/sharp.js:120
  throw new Error(help.join('\n'));
        ^
Error: Could not load the "sharp" module using the linux-x64 runtime
ERR_DLOPEN_FAILED: libvips-cpp.so.8.17.3: cannot open shared object file: No such file or directory
...
Error: Command failed with exit code 1
```

So it is not a cosmetic hook warning — the whole `sync` dies, so background memory sync silently stops working on affected hosts.

## Root cause

The prebuilt binary and its bundled libvips end up in non-adjacent `node_modules` trees:

```
node_modules/@img/sharp-linux-x64/lib/sharp-linux-x64.node          <-- hoisted to top level
node_modules/sharp/node_modules/@img/sharp-libvips-linux-x64/lib/
    libvips-cpp.so.8.17.3                                           <-- left nested under sharp/
```

The binary needs libvips and its `RPATH` only looks in `@img`-adjacent locations:

```
$ readelf -d node_modules/@img/sharp-linux-x64/lib/sharp-linux-x64.node | grep -E 'NEEDED|RPATH'
 (NEEDED)  Shared library: [libvips-cpp.so.8.17.3]
 (RPATH)   Library rpath: [$ORIGIN/../../sharp-libvips-linux-x64/lib:
                           $ORIGIN/../../../sharp-libvips-linux-x64/1.2.4/lib:
                           $ORIGIN/../../node_modules/@img/sharp-libvips-linux-x64/lib:
                           $ORIGIN/../../../node_modules/@img/sharp-libvips-linux-x64/lib:
                           $ORIGIN/../../../../../@img-sharp-libvips-linux-x64-npm-1.2.4-105fd6d44d/node_modules/@img/sharp-libvips-linux-x64/lib]
```

With `$ORIGIN = node_modules/@img/sharp-linux-x64/lib`, the first entry resolves to `node_modules/@img/sharp-libvips-linux-x64/lib` — which does not exist, because that package is under `node_modules/sharp/node_modules/@img/` instead. None of the other entries reach the nested copy either, so `dlopen` fails.

Note `cli/install-check.js` already documents sharp as an optional/OS-specific external ("Excludes optional / OS-specific externals (sharp, fsevents) — missing those [is OK]"), but the runtime path through `@huggingface/transformers` **hard-`require`s** sharp at module load, so that "optional" contract is not honored at runtime — a missing/broken sharp takes down `sync` entirely.

## Reproduction

On a Linux x64 host without system libvips, after a fresh plugin install where npm produces the split layout above:

```
node cli/episodic-memory.js sync     # exits 1 with the ERR_DLOPEN_FAILED trace
```

## Impact

- Background episodic-memory sync crashes on every session start on affected hosts → memory index goes stale, semantic search degrades, and the user only sees a non-blocking "SessionStart hook error".

## Workaround (verified)

Point the RPATH-expected location at the nested bundled libvips:

```sh
cd node_modules/@img
ln -sfn ../sharp/node_modules/@img/sharp-libvips-linux-x64 sharp-libvips-linux-x64
```

After this, `require('sharp')` loads (`sharp.versions.vips === '8.17.3'`) and `sync` runs. A plugin update re-creates the broken layout, so the workaround has to be re-applied.

## Suggested fixes (either or both)

1. **Honor the "sharp is optional" contract at runtime.** Lazy-`require` sharp / the transformers image path and wrap it in try/catch so a failed sharp load degrades gracefully (text embeddings keep working) instead of crashing `sync`. This matches what `install-check.js` already assumes.
2. **Repair the native layout at install time.** Extend `scripts/postinstall.js` to verify `require('sharp')` and, on failure, either `npm rebuild sharp` or create the `@img/sharp-libvips-*` symlink when the bundled copy exists but is only present nested under `sharp/node_modules/@img/`. This mirrors the existing `better-sqlite3` rebuild recovery.

Happy to open a PR for whichever direction you prefer (I'd lean toward #1 as the robust fix, #2 as belt-and-suspenders).

marvinbaudach:
## What & why

Fixes #135. On hosts where `sharp`'s native binding can't `dlopen` libvips, `@huggingface/transformers` (which pulls in sharp transitively) throws at load. Because `src/embeddings.ts` imported transformers at the **top level**, that failure crashed *every* consumer at import time — including the `SessionStart` `sync --background` hook, which exited 1 with a raw stack trace and silently stopped background sync.

Note that `cli/install-check.js` already documents sharp as an optional/OS-specific external ("missing those is OK"), but the runtime path did not honor that contract. This PR makes it real: a missing/broken sharp degrades semantic features instead of taking the process down.

## Changes

- **`src/embeddings.ts`** — load `@huggingface/transformers` lazily inside `initEmbeddings()` (via dynamic `import()`), wrapped in try/catch. On failure it throws a typed **`EmbeddingsUnavailableError`** (with `cause`). Merely importing `embeddings.ts` no longer touches sharp; the failure is now catchable and confined to callers that actually embed. The `env.allowLocalModels` / `env.useBrowserCache` config moves into the lazy init (still applied before the first `pipeline()` call).
- **`src/sync.ts`** — in the indexing block, catch an embedding-backend load failure, push a clear actionable entry to `result.errors`, log a one-line diagnostic to stderr, and **skip semantic indexing for that run** instead of throwing out of `syncConversations`. Copying and summaries still complete.

The failure stays **visible** (stderr + `result.errors`) — this is not error-hiding (cf. #94); it just no longer crashes the whole sync. The embedding-migration phase in `sync-cli.ts` was already wrapped in try/catch, so it benefits automatically once the import is lazy.

Out of scope: `search.ts` / `indexer.ts` still call `initEmbeddings()` without a catch, so on an affected host those commands now surface a clean typed `EmbeddingsUnavailableError` instead of a crash-at-import — happy to add fallback-to-FTS there in a follow-up if you'd like.

## Testing

- `test/embeddings-unavailable.test.ts` — `initEmbeddings()` rejects with `EmbeddingsUnavailableError` (typed, actionable, preserves cause) when the transformers load fails.
- `test/sync-embeddings-unavailable.test.ts` — `syncConversations` **resolves** (does not throw) when the embedder can't load: the file is still copied, `indexed === 0`, and a clear `(embeddings)` error is recorded.
- Full suite: `tsc --noEmit` clean; **209/210 tests pass**. The single failure (`test/show.test.ts` markdown formatting) reproduces on unmodified `main` and is unrelated to this change.

🤖 Prepared with Claude Code

### DIFF
--- a/src/embeddings.ts
+++ b/src/embeddings.ts
@@ -1,9 +1,4 @@
-import { pipeline, FeatureExtractionPipeline, env } from '@huggingface/transformers';
-
-// Disable progress callbacks to prevent stdout pollution in MCP context
-// In MCP, stdout is reserved for JSON-RPC communication.
-env.allowLocalModels = true;
-env.useBrowserCache = false;
+import type { FeatureExtractionPipeline } from '@huggingface/transformers';
 
 /**
  * Embedding model configuration.
@@ -23,16 +18,54 @@ export const BGE_QUERY_PREFIX = 'Represent this sentence for searching relevant 
 
 let embeddingPipeline: FeatureExtractionPipeline | null = null;
 
-export async function initEmbeddings(): Promise<void> {
-  if (!embeddingPipeline) {
-    console.error('Loading embedding model (first run may take time)...');
-    embeddingPipeline = await pipeline(
-      'feature-extraction',
-      MODEL_ID,
-      { dtype: MODEL_DTYPE, progress_callback: () => {} }
-    );
-    console.error('Embedding model loaded');
+/**
+ * Thrown when the embedding backend can't be loaded. The most common cause is
+ * that `@huggingface/transformers` eagerly requires `sharp`, whose native
+ * binding fails to `dlopen` libvips on some hosts (#135) — even though sharp is
+ * only needed for image inputs, not the text feature-extraction this package
+ * uses. Callers that can proceed without semantic features (e.g. background
+ * sync indexing) should catch this and degrade gracefully rather than crash.
+ */
+export class EmbeddingsUnavailableError extends Error {
+  constructor(message: string, options?: { cause?: unknown }) {
+    super(message, options);
+    this.name = 'EmbeddingsUnavailableError';
   }
+}
+
+export async function initEmbeddings(): Promise<void> {
+  if (embeddingPipeline) return;
+
+  // Load @huggingface/transformers lazily. Its module graph eagerly requires
+  // `sharp`, so a static top-level import would crash *every* consumer of this
+  // file at import time on hosts where sharp's native binding can't load —
+  // including code paths that never embed anything. Importing here keeps the
+  // failure catchable and confined to callers that actually need embeddings.
+  let pipeline: typeof import('@huggingface/transformers').pipeline;
+  try {
+    const transformers = await import('@huggingface/transformers');
+    // Disable progress callbacks / remote cache to prevent stdout pollution in
+    // MCP context, where stdout is reserved for JSON-RPC communication.
+    transformers.env.allowLocalModels = true;
+    transformers.env.useBrowserCache = false;
+    pipeline = transformers.pipeline;
+  } catch (error) {
+    throw new EmbeddingsUnavailableError(
+      'Failed to load the embedding backend (@huggingface/transformers). This ' +
+      'usually means the native "sharp" binding could not load libvips; ' +
+      'semantic search and indexing are unavailable until it is fixed. ' +
+      `Underlying error: ${error instanceof Error ? error.message : String(error)}`,
+      { cause: error }
+    );
+  }
+
+  console.error('Loading embedding model (first run may take time)...');
+  embeddingPipeline = await pipeline(
+    'feature-extraction',
+    MODEL_ID,
+    { dtype: MODEL_DTYPE, progress_callback: () => {} }
+  );
+  console.error('Embedding model loaded');
 }
 
 export async function generateEmbedding(text: string): Promise<number[]> {
--- a/src/sync.ts
+++ b/src/sync.ts
@@ -144,43 +144,68 @@ export async function syncConversations(
 
   // Index copied files (unless skipIndex is set)
   if (!options.skipIndex && filesToIndex.length > 0) {
-    const { initDatabase, insertExchange } = await import('./db.js');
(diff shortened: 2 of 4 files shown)


obra:
Thanks @marvinbaudach! Merged to main as `7e0a415` (squashed as part of a 13-PR batch, with dist rebuilt on top). This makes sync degrade gracefully when the embedding backend can't load, so SessionStart no longer crashes and copy+summaries keep working.

PR #132, opened 2026-07-19
fix(indexer): skip vanished transcripts instead of aborting

arimu1:
## Summary
- Wrap archive copy + `parseConversation` in try/catch for the bulk index, single-session, and unprocessed paths.
- On read failure, log `Skipped <file> (read failed: …)` and continue.

Fixes #121

## Test plan
- [ ] Manually delete a listed transcript mid-`index` (or simulate ENOENT) and confirm the run completes for remaining files

Made with [Cursor](https://cursor.com)

Issue #121, opened 2026-07-05
index-all aborts entirely when a source transcript vanishes mid-run (ENOENT race with Claude Code's transcript cleanup)


taehyeonglim:
`episodic-memory index` lists source files, then parses each one. If a transcript is deleted between the listing and the read, `parseConversation` throws ENOENT and the **whole run dies** instead of skipping the file:

```
Error: ENOENT: no such file or directory, open '~/.claude/projects/<project>/<uuid>.jsonl'
Error: Command failed with exit code 1
```

This is a real race on live corpora: Claude Code's ~30-day transcript cleanup (and ordinary session deletion) can remove files while a long index run is in flight. We hit it in production on a ~4k-conversation corpus (v1.4.2 `1075769`, macOS 15, node 26).

Per-file try/catch — skip loudly, keep going — fixes it. Patch we're running in production:

```diff
diff --git a/src/indexer.ts b/src/indexer.ts
index b3525a1..d7b9a64 100644
--- a/src/indexer.ts
+++ b/src/indexer.ts
@@ -106,15 +106,23 @@ export async function indexConversations(
       const sourcePath = path.join(projectPath, file);
       const archivePath = path.join(projectArchive, file);
 
-      // Copy to archive (ensure parent dirs exist for subagent files)
-      if (!fs.existsSync(archivePath)) {
-        fs.mkdirSync(path.dirname(archivePath), { recursive: true });
-        fs.copyFileSync(sourcePath, archivePath);
-        console.log(`  Archived: ${file}`);
-      }
+      // [NERV patch] Source transcripts can vanish mid-run (Claude Code's
+      // ~30-day cleanup). Skip the file loudly instead of aborting the run.
+      let exchanges;
+      try {
+        // Copy to archive (ensure parent dirs exist for subagent files)
+        if (!fs.existsSync(archivePath)) {
+          fs.mkdirSync(path.dirname(archivePath), { recursive: true });
+          fs.copyFileSync(sourcePath, archivePath);
+          console.log(`  Archived: ${file}`);
+        }
 
-      // Parse conversation
-      const exchanges = await parseConversation(sourcePath, project, archivePath);
+        // Parse conversation
+        exchanges = await parseConversation(sourcePath, project, archivePath);
+      } catch (error) {
+        console.log(`  Skipped ${file} (read failed: ${error instanceof Error ? error.message : error})`);
+        continue;
+      }
 
       if (exchanges.length === 0) {
[...]

PR #143, opened 2026-07-31
fix: skip dangling project symlinks instead of aborting directory walks

### LINKED ISSUE #142: One dangling symlink in ~/.claude/projects crashes sync, index, verify, and rebuild (ENOENT from bare statSync in every directory walk)
## What happens

Every directory walker in the codebase stats each `readdirSync` entry directly:

```ts
const projectPath = path.join(sourceDir, project);
const stat = fs.statSync(projectPath);   // throws ENOENT on a dangling symlink
```

If `~/.claude/projects` (or the archive dir) contains a symlink whose target no longer exists, `statSync` throws `ENOENT` and the entire run aborts — no remaining project gets synced, indexed, or verified.

Affected call sites at v1.4.2 (`1075769`):

- `src/sync.ts:105` — `syncConversations` (runs on every SessionStart)
- `src/indexer.ts:79` — `indexConversations`
- `src/indexer.ts:207` — `indexSession`
- `src/indexer.ts:307` — `indexUnprocessed`
- `src/verify.ts:48` — `verifyIndex`
- `src/index-cli.ts:95` — `rebuild` summary-deletion walk

## How I hit it

A storage migration moved project data and left a symlink in `~/.claude/projects` pointing at a target that no longer exists. From that moment, every SessionStart sync and every `index-all` / `index-cleanup` / `verify` run died with the same ENOENT before touching any other project. Episodic memory was fully down until the entry was found and removed by hand.

This is the same "one bad entry aborts the whole run" failure family as #121 (transcript files vanishing mid-run, fixed by #132), one directory level up: project entries instead of transcript files.

## Minimal repro

```bash
mkdir -p /tmp/em-repro/projects/real-project /tmp/em-repro/archive
printf '{"type":"user","message":{"role":"user","content":"hello"}}\n' \
  > /tmp/em-repro/projects/real-project/aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee.jsonl
ln -s /nonexistent-target /tmp/em-repro/projects/migrated-away

# from a checkout of this repo
node --input-type=module -e "
import { syncConversations } from './dist/sync.js';
console.log(await syncConversations('/tmp/em-repro/projects', '/tmp/em-repro/archive', { skipIndex: true, skipSummaries: true }));
"
```

Result at v1.4.2:

```
Error: ENOENT: no such file or directory, stat '/tmp/em-repro/projects/migrated-away'
    at Object.statSync (node:fs:1746:25)
    at syncConversations (.../dist/sync.js:76:25)
```

The healthy `real-project` sibling is never synced.

## Why "just skip symlinks" is the wrong fix

Valid symlinked project directories exist in the wild (I have two in my own `~/.claude/projects`) and must keep working. So `readdirSync(..., { withFileTypes: true })` / `lstatSync` dirent checks — which would stop following symlinks entirely — would trade a crash for silent data loss. The walk has to keep `statSync` follow semantics and tolerate the ENOENT.

## Environment

- episodic-memory v1.4.2 (`1075769`); originally hit on the superpowers-marketplace install of v1.0.15 — the call sites are unchanged between the two
- macOS (darwin arm64), Node v22.23.2

I have a fix with tests ready and will open a PR.

WVenerable83:
Fixes #142

## Problem

A dangling symlink in `~/.claude/projects` (for example, left behind by a storage migration) crashes every walker that stats `readdirSync` entries: `syncConversations`, `indexConversations`, `indexSession`, `indexUnprocessed`, `verifyIndex`, and the `rebuild` summary-deletion walk all die with `ENOENT` before processing any remaining project. Since sync runs on SessionStart, one stale symlink takes episodic memory fully down.

```
Error: ENOENT: no such file or directory, stat '.../projects/migrated-away'
    at Object.statSync (node:fs:1746:25)
    at syncConversations (.../dist/sync.js:76:25)
```

Same failure family as #121 / #132 (transcripts vanishing mid-run), one directory level up.

## Fix

Add `statIfExists()` to `paths.ts` — a `statSync` that returns `null` instead of throwing when the entry can't be stat'ed — and use it at all six walk sites:

```ts
const stat = statIfExists(projectPath);
if (!stat?.isDirectory()) continue;
```

Design notes:

- **Valid symlinks keep working.** The helper keeps `statSync` follow semantics (not `lstat`), so symlinked project directories are still traversed; only entries whose target is gone are skipped. The new sync test asserts a valid-symlinked project still syncs.
- **Skip is silent**, matching the existing silent skip of non-directory entries one line down. Happy to add a log line if you'd prefer these surfaced.
- An entry deleted between `readdir` and `stat` (the #121-style race, at the project level) is covered by the same guard.

## Tests

New `test/dangling-symlink.test.ts` (follows the `exclude-nested.test.ts` isolation pattern):

- sync copies a real project **and** a valid-symlinked project while skipping a dangling sibling (`copied === 2`, no errors)
- `indexConversations` / `indexUnprocessed` index the real project despite a dangling sibling
- `indexSession` completes a not-found scan without aborting
- `verifyIndex` walks an archive containing a dangling symlink
- `statIfExists` unit: real dir / valid symlink → `Stats`, dangling symlink / missing path → `null`
- the whole suite `describe.skipIf`s on platforms where symlink creation is unavailable (Windows without Developer Mode)

All six tests fail with the exact production `ENOENT` (or missing helper) before the fix and pass after.

Full suite: 212/213 passing locally. The one failure, `test/verify.test.ts > repairIndex > re-indexes outdated files during repair`, is a pre-existing same-millisecond flake — it fails identically on a clean `main` checkout on a fast machine (`expected 1785529540716 to be greater than 1785529540716`). Happy to file it separately.

## Repro

```bash
mkdir -p /tmp/em-repro/projects/real-project /tmp/em-repro/archive
printf '{"type":"user","message":{"role":"user","content":"hello"}}\n' \
  > /tmp/em-repro/projects/real-project/aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee.jsonl
ln -s /nonexistent-target /tmp/em-repro/projects/migrated-away

node --input-type=module -e "
import { syncConversations } from './dist/sync.js';
console.log(await syncConversations('/tmp/em-repro/projects', '/tmp/em-repro/archive', { skipIndex: true, skipSummaries: true }));
"
```

Before: `ENOENT` crash, nothing synced. After: `{ copied: 1, skipped: 0, indexed: 0, summarized: 0, errors: [] }`.

`dist/` is untouched per repo convention (regenerated at release); `npm run build` compiles clean with the change.

### DIFF
--- a/src/index-cli.ts
+++ b/src/index-cli.ts
@@ -2,7 +2,7 @@
 import { verifyIndex, repairIndex } from './verify.js';
 import { indexSession, indexUnprocessed, indexConversations } from './indexer.js';
 import { initDatabase } from './db.js';
-import { getDbPath, getArchiveDir } from './paths.js';
+import { getDbPath, getArchiveDir, statIfExists } from './paths.js';
 import fs from 'fs';
 import path from 'path';
 
@@ -92,7 +92,7 @@ async function main() {
           const projects = fs.readdirSync(archiveDir);
           for (const project of projects) {
             const projectPath = path.join(archiveDir, project);
-            if (!fs.statSync(projectPath).isDirectory()) continue;
+            if (!statIfExists(projectPath)?.isDirectory()) continue;
 
             const summaries = fs.readdirSync(projectPath).filter(f => f.endsWith('-summary.txt'));
             for (const summary of summaries) {
--- a/src/indexer.ts
+++ b/src/indexer.ts
@@ -6,7 +6,7 @@ import { parseConversation } from './parser.js';
 import { initEmbeddings, generateExchangeEmbedding } from './embeddings.js';
 import { summarizeConversation } from './summarizer.js';
 import { ConversationExchange } from './types.js';
-import { getArchiveDir, getExcludedProjects, getConversationSourceDirs, findJsonlFiles } from './paths.js';
+import { getArchiveDir, getExcludedProjects, getConversationSourceDirs, findJsonlFiles, statIfExists } from './paths.js';
 import { formatErrorSentinel, shouldQueueForSummary } from './summary-sentinel.js';
 
 // Set max output tokens for Claude SDK (used by summarizer)
@@ -76,9 +76,9 @@ export async function indexConversations(
     // Skip if limiting to specific project
     if (limitToProject && project !== limitToProject) continue;
     const projectPath = path.join(sourceDir, project);
-    const stat = fs.statSync(projectPath);
+    const stat = statIfExists(projectPath);
 
-    if (!stat.isDirectory()) continue;
+    if (!stat?.isDirectory()) continue;
 
     const files = findJsonlFiles(projectPath, excludedDirSet);
 
@@ -204,7 +204,7 @@ export async function indexSession(sessionId: string, concurrency: number = 1, n
     if (excludedProjects.includes(project)) continue;
 
     const projectPath = path.join(sourceDir, project);
-    if (!fs.statSync(projectPath).isDirectory()) continue;
+    if (!statIfExists(projectPath)?.isDirectory()) continue;
 
     const files = findJsonlFiles(projectPath, excludedDirSet).filter(f => f.includes(sessionId));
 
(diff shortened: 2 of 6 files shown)


obra:
Thanks @WVenerable83! Merged to main as `cff1946` (squashed as part of a 13-PR batch, with dist rebuilt on top). This stops a single dangling project symlink from crashing sync, index, verify, and rebuild.

PR #144, opened 2026-08-01
fix(search): multi-word text search as AND of token LIKEs

### LINKED ISSUE #127: Text search matches the whole query as one literal substring — multi-word queries return nothing (mode=text and the text half of both)
**Version:** 1.4.2 (code unchanged on current `main`). Affects `--mode text` and the text half of the default `both` mode — CLI and MCP `search` both.

## Symptom

Any multi-word query that isn't a verbatim phrase from a conversation returns nothing in text mode. The words may each appear thousands of times; unless they occur *contiguously, in that order, in one message*, the text branch finds 0 rows.

Measured on a 22,366-exchange index — per-word vs whole-phrase hits for the query "date filter empty results":

| needle | exchanges matching |
|---|---|
| `date` | 6,905 |
| `filter` | 1,960 |
| `empty` | 1,479 |
| `results` | 1,507 |
| `date filter empty results` (what the code actually searches) | 0 |

## Mechanism

The text branch (src/search.ts:187, 194) binds the whole query as one `LIKE` pattern:

```ts
WHERE (e.user_message LIKE ? OR e.assistant_message LIKE ?)
// ...
const textResults = textStmt.all(`%${query}%`, `%${query}%`, ...);
```

So the entire query string must occur as a single contiguous substring. Multi-word queries are almost never verbatim quotes of the conversation, so in practice text mode only works for single words and exact phrases.

Side effect on `both` (the default): the text branch can't act as a fallback when the vector branch starves (#126) — both halves return empty simultaneously, which is why #126 presents as "No results found" even in the default mode.

(Also worth fixing while in there: the query is bound into the LIKE pattern without escaping `%`/`_`, so a query containing `%` silently changes matching semantics.)

## Repro

1. Index a conversation containing "I fixed the authentication bug in the login flow".
2. `search "authentication login" --mode text` → no results (both words present, not adjacent).
3. `search "authentication bug" --mode text` → found (happens to be a verbatim substring).

## Fix options

Two levels, not mutually exclusive:

**Option A — split into AND-ed terms (no migration, a few lines).** Tokenize the query on whitespace and require every term to match somewhere in the exchange:

```sql
WHERE (e.user_message LIKE '%date%' OR e.assistant_message LIKE '%date%')
  AND (e.user_message LIKE '%filter%' OR e.assistant_message LIKE '%filter%')
  -- ... one clause per term
```

All words, any order, either field. Keeps the existing recency ordering (`ORDER BY timestamp DESC`), no schema change, no re-index. No relevance ranking and no stemming — but it turns "returns nothing" into "returns the right rows", and it restores the text branch as a working fallback for `both`.

**Option B — FTS5 (migration).** A proper full-text index over the message columns:

- A `fts5` virtual table populated from `exchanges`, kept in sync by the indexer on insert/delete (or by triggers).
- Gains: BM25 relevance ranking (instead of recency-only), quoted phrase queries, prefix matching, proper tokenization.
- Costs: a schema migration for existing installs — same shape as the `embedding_version` re-embed flow but much cheaper, since an FTS rebuild is pure SQLite (one-time `INSERT INTO ... SELECT` on upgrade, no model involved); and user-supplied query strings need escaping into valid FTS5 syntax so `"`, `-`, `OR` etc. don't error.

A pragmatic path is A now (tiny, unbreaks `both`), B if/when text search should be a first-class surface rather than a fallback. Happy to send a PR for either.

arimu1:
## Summary

Text mode (and the text half of default `both`) bound the entire query as a single `LIKE '%query%'` pattern. Multi-word queries only matched when those words appeared as one contiguous substring in a message — typical searches like `authentication login` returned nothing even when both words were present.

This implements issue #127 option A:

- Tokenize the query on whitespace
- Require every term to match somewhere in `user_message` or `assistant_message` (AND of per-token LIKE, either field, any order)
- Escape `%` / `_` (and `\`) in user terms so they are treated as literals (`ESCAPE '\'`)

No schema change, no re-index. Recency ordering (`ORDER BY timestamp DESC`) unchanged. Restores text search as a working fallback when the vector branch is empty or starved.

## Test plan

- [x] `npm test -- test/text-search-multiword.test.ts` (8 tests)
- [x] `npm test -- test/search-metadata-filters.test.ts test/cross-harness-recall.test.ts` (no regression)
- [x] Covered cases:
  - non-contiguous multi-word match (`authentication login` on “I fixed the authentication bug in the login flow”)
  - single-word still works
  - AND semantics (missing term excludes the row)
  - terms split across user vs assistant messages
  - `%` / `_` treated as literals

## AI assistance

Implemented with AI assistance (Grok / Cursor harness). Human-reviewed for scope and test coverage.

Fixes #127


obra:
Thanks @arimu1! Merged to main as `4248fe3` (squashed as part of a 13-PR batch, with dist rebuilt on top). This matches multi-word text queries as an AND of token substrings instead of one literal substring, so multi-word searches return results.

PR #145, opened 2026-08-01
fix(summarizer): surface SDK result text and fail-fast on auth errors

### LINKED ISSUE #138: CLI auth failure (401) is logged as the contradictory "Summarizer SDK error: success" with the real error text discarded, and sync has no fail-fast, so each run burns ~32 minutes on a batch that is doomed by the same auth error
**Version:** episodic-memory 1.4.2 (superpowers-marketplace), macOS (arm64), Claude Code CLI authenticated via subscription OAuth (no `ANTHROPIC_API_KEY` set).

Two related defects, found while debugging why every sync on this machine reported ~10 summary failures. Root cause turned out to be expired Claude CLI OAuth credentials, but the plugin's error handling actively pointed away from that, and each sync spent about half an hour failing.

## Defect 1: error opacity, the actual error text is discarded

`callClaude()` in `src/summarizer.ts` throws `SummarizerSdkError(subtype, session_id)` whenever the SDK result message has `is_error: true`:

```ts
// src/summarizer.ts (v1.4.2), inside callClaude()
if ((message as any).is_error) {
  throw new SummarizerSdkError((message as any).subtype || 'unknown', (message as any).session_id);
}
```

But when the CLI fails auth, the result message looks like this (reproduced directly with the same resume call the summarizer makes):

```
$ claude -p --resume <session-id> --output-format json
{
  "type": "result",
  "subtype": "success",
  "is_error": true,
  "result": "Failed to authenticate. API Error: 401 ... \"OAuth access token has expired.\" ...",
  ...
}
```

So `subtype` is `"success"` while `is_error` is `true`, and the only field that says what actually went wrong (`result`) is thrown away. The user-visible output becomes the self-contradictory:

```
Summary generation failed: Summarizer SDK error: success (session 3beec7df-2ec4-45af-bb65-a96b4bddf734)
...
Check your API configuration (EPISODIC_MEMORY_API_BASE_URL / ANTHROPIC_API_KEY).
```

That hint sent me down the wrong path entirely (checking API base URL and key config) when the real fix was `claude` CLI re-authentication. Nothing in the log ever shows the words "401", "authenticate", or "OAuth".

**Suggested fix:** when the result message carries a string `result`, include it (truncated, say 200 to 300 chars) in the `SummarizerSdkError` message alongside the subtype. That single change would have made the log say "OAuth access token has expired" instead of "success".

## Defect 2: no fail-fast, one expired token costs ~32 minutes per sync

Each failed headless call takes about 190 seconds, because the CLI retries the 401 with backoff before giving up. Sync then works through its full batch sequentially:

```ts
// src/sync.ts (v1.4.2)
const summaryLimit = options.summaryLimit ?? 10;   // default batch of 10
const toSummarize = filesToSummarize.slice(0, summaryLimit);
...
for (const { path: filePath, sessionId } of toSummarize) {   // sequential
```

10 x ~190s means roughly 32 minutes of a background sync hammering an endpoint that will reject every call for the same reason. And since auth failures leave no sentinel, the same batch retries on the next sync (that part overlaps #103, see below).

**Suggested fix:** authentication errors are global, not per-conversation. After the first result whose error text indicates an auth failure (401 / "Failed to authenticate" / `authentication_error`), skip the rest of the summarization batch for that run, mark the skipped items retryable via the existing `__ERRORED__` sentinel machinery (`src/summary-sentinel.ts`), and print one clear message: "Claude CLI authentication failed, run `claude` and re-authenticate, summaries will retry on the next sync."

## Relationship to existing issues

- #103 covers the same `is_error` + `subtype: "success"` throw path but focuses on the missing sentinel / infinite re-queue / billing. This issue is about the two things #103 doesn't cover: the discarded `result` error text (defect 1) and the lack of batch-level fail-fast on auth errors (defect 2). Fixing #103's sentinel gap would not fix either of these.
- #96 (error sentinels) and #87 (reentrancy) are adjacent lineage but different defects.
- #94 (invisible SessionStart failures) compounds defect 1: the misleading message is often the only thing a user ever finds in the log.

arimu1:
## Summary

Fixes #138.

When the Claude CLI fails auth, the Agent SDK still returns `subtype: "success"` with `is_error: true`, and the real explanation only lives in the `result` string (e.g. `401` / `OAuth access token has expired`). `callClaude` previously threw `SummarizerSdkError` with only the subtype, so logs said the contradictory **Summarizer SDK error: success** and pointed people at API base URL / key config instead of re-auth.

This PR:

1. **Surfaces SDK `result` text** on `SummarizerSdkError` (truncated to 300 chars) so logs include the real auth/API failure.
2. **Fail-fasts the sync summary batch** after the first global auth failure (`isAuthFailure`), writing `__ERRORED__` sentinels for remaining items so a single expired OAuth token no longer burns ~32 minutes (10 × ~190s CLI retries).

## Changes

- `src/summarizer.ts`: optional `detail` on `SummarizerSdkError`; pass `message.result` from `is_error` results; export `isAuthFailure` / `truncateSdkErrorDetail`
- `src/sync.ts`: after first auth failure, skip rest of batch with error sentinels + clear re-auth message
- Tests for opaque `subtype: "success"` + result text, auth fail-fast, and non-auth errors still processing the full batch
- `dist/` rebuilt (`summarizer` / `sync` / stale `version` 1.4.1→1.4.2)

## Test plan

- [x] `npm test -- test/summarizer-options.test.ts test/summarizer-resume-fallback.test.ts test/sync-error-sentinel.test.ts` (52 passed)
- [ ] Manual: expire/revoke Claude CLI OAuth, run `episodic-memory sync` with several unsummarized sessions — first failure should mention OAuth/401, remaining batch skipped with sentinels, no multi-minute hammering

## AI disclosure

Assisted by **Grok** (xAI) in the Cursor harness as implementator for arimu1. Human-directed issue selection and review.


obra:
Thanks @arimu1! Merged to main as `f06748a`. Folded into the consolidated summarizer pass: SDK result text is surfaced and auth failures fail-fast the batch instead of burning ~32 minutes on doomed calls.

PR #147, opened 2026-08-02
fix: fall back to transcript summarization when the resume subprocess exits nonzero

### LINKED ISSUE #146: Summarization permanently fails for background-agent sessions — resume subprocess exit bypasses the fallback, stderr is discarded
Summaries for some conversations fail on every retry with:

```
Summary generation failed: Claude Code process exited with code 1
```

**Root cause (two compounding problems):**

1. For conversations ≤15 exchanges the summarizer resumes the original session (`resume: sessionId`). Sessions registered as **background agents** make `claude --resume` exit 1. With the SDK's `stderr` callback wired up, the subprocess says:

   ```
   Error: Session <id> is currently running as a background agent (bg).
   Use `claude agents` to find and attach to it, or add --fork-session to branch off a copy.
   ```

2. `summarizeConversation` already has a retry-without-resume fallback, but it only fires for `isResumeFailure()` (SummarizerSdkError subtype `error_during_execution`). A subprocess crash is thrown by the SDK as a **plain `Error`**, so the fallback never fires. The `__ERRORED__` sentinel (#96) then re-queues it hourly and it fails identically forever. Because the SDK spawns the CLI with stderr ignored (no `stderr` callback passed), the sentinel records only the opaque exit code, making this hard to diagnose.

Environment: Windows 11, Claude Code 2.1.220, episodic-memory 1.4.2. Reproduced deterministically against a real bg-agent session; verified the fallback fix produces a summary on the next sync.

Related: #96 (error sentinels), #87 (reentrancy guard), #83 (persistSession).

PR with fix + tests incoming.

jolive87:
Fixes #146.

- Add `isProcessExitFailure()` — matches the SDK's plain-`Error` throws (`exited with code` / `terminated by signal`) — and include it in the retry-without-resume condition alongside `isResumeFailure()`. Background-agent sessions (which `claude --resume` refuses with exit 1) now summarize via the transcript-text path instead of erroring out permanently.
- Capture the subprocess's stderr via the SDK's `stderr` callback in `callClaude` and append it to process-exit errors, so `__ERRORED__` sentinels record the actual cause (e.g. the bg-agent refusal message) instead of an opaque exit code.
- Tests: two new cases in `test/summarizer-resume-fallback.test.ts` (fallback fires on subprocess exit; stderr is appended to the surfaced error). All existing summarizer/sentinel tests pass.

An alternative considered was passing `forkSession: true` when resuming, which would keep resume-quality summaries for bg sessions — happy to switch to that if you prefer; the fallback fix is strictly more defensive (it covers any future exit-1 cause, not just bg sessions).

Note: this also surfaced a quality quirk worth a separate look — on the transcript-text path, haiku sometimes replies *to* the conversation instead of summarizing it (the `extractSummary` fallback then stores the raw reply). Not addressed here.

### DIFF
--- a/src/summarizer.ts
+++ b/src/summarizer.ts
@@ -31,6 +31,17 @@ export function isResumeFailure(error: unknown): boolean {
   return error instanceof SummarizerSdkError && error.subtype === 'error_during_execution';
 }
 
+/**
+ * True when the SDK's spawned Claude Code subprocess died before returning a
+ * result (e.g. `--resume` exiting 1 because the session is a background agent
+ * that can't be resumed without --fork-session). The SDK throws these as plain
+ * Errors, not SummarizerSdkError, so isResumeFailure never matches them —
+ * callers that resumed a session should fall back to the non-resume path.
+ */
+export function isProcessExitFailure(error: unknown): boolean {
+  return error instanceof Error && /exited with code|terminated by signal/.test(error.message);
+}
+
 export interface CodexSummarizerCommand {
   command: string;
   args: string[];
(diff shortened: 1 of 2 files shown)


obra:
Thanks @jolive87! Merged to main as `f06748a`. Folded into the consolidated summarizer pass: a nonzero-exit resume now falls back to transcript summarization, with subprocess stderr captured.

PR #149, opened 2026-08-06
fix: stop summarizer subprocesses from launching the user's MCP servers

lvillar:
## Problem

Each summarization call spawns a Claude subprocess via the Agent SDK (`query()` in `summarizer.ts`). That subprocess inherits the user's full configuration: every MCP server in `~/.claude.json` plus all enabled plugins and their MCP servers.

On an MCP-heavy install this is brutal. My setup has ~20 configured MCP servers (several database servers, puppeteer, playwright, context7...). With 10 concurrent summarizations, the background sync fanned out to **~200 node processes** — load average 59, swap exhausted — starving everything else on the machine. It looked enough like malware that I audited it before finding the real cause. The reentrancy guard (#87) prevents the recursive cascade, but each summarizer still drags the whole MCP fleet with it.

## Fix

Summarizers only write text — they need no tools, plugins, or settings:

- `strictMcpConfig: true` → the SDK passes `--strict-mcp-config`, so the user's and project's MCP configs are ignored (no `mcpServers` are passed, so the subprocess gets none)
- `settingSources: []` → no user settings, plugins, or hooks are loaded into the subprocess

## Testing

- Two new cases in `test/summarizer-options.test.ts`; the file passes (27/27) and `tsc --noEmit` is clean
- Verified on a live install: patched summarizer subprocesses carry `--strict-mcp-config` on their command line and spawn 1 child process instead of ~7 MCP servers each; a full sync of a 370-conversation backlog ran without measurable load
- `dist/summarizer.js` rebuilt and committed per the repo convention

### DIFF
--- a/dist/summarizer.js
+++ b/dist/summarizer.js
@@ -94,6 +94,15 @@ export function buildSummarizerQueryOptions(args) {
         env: getApiEnv(),
         resume: sessionId,
         persistSession: false,
+        // Summarizers never call tools, so skip the user's MCP config entirely.
+        // Without this, every summarization subprocess launches all configured MCP
+        // servers; with concurrent summarizations that fans out to hundreds of
+        // processes on MCP-heavy installs.
+        strictMcpConfig: true,
+        // strictMcpConfig only blocks ~/.claude.json MCP servers; user settings
+        // still load every enabled plugin (each with its own MCP server) into the
+        // subprocess. Summarizers need no plugins, hooks, or user settings at all.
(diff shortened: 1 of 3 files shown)


obra:
Thanks @lvillar! Merged to main as `58363b2` (squashed as part of a 13-PR batch, with dist rebuilt on top). This isolates the summarizer subprocess from your MCP servers and settings (settingSources:[] + strictMcpConfig), closing the summarizer-hijack class of bug.

Issue #105, opened 2026-06-01
v1.4.2: onnxruntime-common not hoisted on Linux/WSL2 — clean-install repro of #95 Bug 2 (conflicting versions confirmed)


jtarmstrong:
## Summary

On a clean Linux (WSL2) install of `episodic-memory@1.4.2`, `onnxruntime-common` is **not** hoisted to top-level `node_modules/`. It is physically nested under **both** `onnxruntime-node` and `onnxruntime-web`, each carrying a *different* version, so npm cannot hoist either. `@huggingface/transformers`'s bare `import { Tensor } from "onnxruntime-common"` then fails with `ERR_MODULE_NOT_FOUND`.

This is a clean-install reproduction of **Bug 2 from #95**, which was closed as "does NOT reproduce — onnxruntime-common IS hoisted on clean installs (macOS + Windows 11)." It reproduces deterministically here on **Linux**, so it is not the Windows-postinstall red herring suspected in that thread — it's a genuine cross-platform npm hoisting failure driven by conflicting `onnxruntime-common` versions.

## Environment

- Plugin: `episodic-memory@1.4.2` (via `superpowers-marketplace`)
- OS: Linux 6.6.87.2-microsoft-standard-WSL2 x86_64 (WSL2)
- Node: v22.22.1
- npm: 10.9.4

## Evidence

**1. `onnxruntime-common` is not at top-level — it's nested under both parents:**

```
$ ls node_modules/onnxruntime-common
ls: cannot access 'node_modules/onnxruntime-common': No such file or directory

$ find node_modules -name onnxruntime-common -type d
node_modules/onnxruntime-web/node_modules/onnxruntime-common
node_modules/onnxruntime-node/node_modules/onnxruntime-common
```

**2. The two parents pull conflicting versions — this is why npm won't hoist either:**

```
node_modules/onnxruntime-web/node_modules/onnxruntime-common   -> 1.24.0-dev.20251116-b39e144322
node_modules/onnxruntime-node/node_modules/onnxruntime-common  -> 1.24.3
```

`@huggingface/transformers` declares only the two parents, not `onnxruntime-common` itself:

```json
"onnxruntime-node": "1.24.3",
"onnxruntime-web": "1.26.0-dev.20260416-b7804b056c"
```

**3. The bare import fails (both CJS resolve and the real ESM entry point):**

```
$ node -e "require.resolve('onnxruntime-common')"
Error: Cannot find module 'onnxruntime-common'

[...]

Issue #95, opened 2026-05-18
Plugin fails to load on Windows: two upstream dep-resolution bugs slip past mcp-server-wrapper.js


danielsimonjr:
# Plugin fails to load on Windows: two upstream dep-resolution bugs slip past `mcp-server-wrapper.js`

## Summary

`episodic-memory@1.4.1` fails to load as a Claude Code plugin MCP server on a fresh Windows install. Two distinct failures, both triggered by `npm install`-time gaps that the wrapper's startup check doesn't detect:

1. **Partial `node_modules` extraction** leaves `better-sqlite3` missing its `lib/` and native binding. The wrapper's `existsSync(node_modules)` check sees the corrupt folder, declares deps healthy, and the server crashes on import.
2. **After running `npm install` manually**, `@huggingface/transformers@4.2.0` fails to resolve a bare `import { Tensor } from "onnxruntime-common"` because that package isn't declared as a direct dep and isn't hoisted to top-level.

Both reproduce deterministically. Both have small upstream fixes.

## Environment

- Plugin: `episodic-memory@1.4.1` (installed via `superpowers-marketplace`)
- OS: Windows 11
- Node: v24.15.0
- npm: 11.12.1
- Plugin path: `C:\Users\danie\.claude\plugins\cache\superpowers-marketplace\episodic-memory\1.4.1\`

### Symptom

```
Error: Cannot find package
'C:\Users\danie\.claude\plugins\cache\superpowers-marketplace\episodic-memory\1.4.1
\node_modules\better-sqlite3\index.js' imported from
...\1.4.1\dist\mcp-server.js
    at legacyMainResolve (node:internal/modules/esm/resolve:201:26)
    ...
  code: 'ERR_MODULE_NOT_FOUND'
```

### State on disk

```
node_modules/better-sqlite3/
├── deps/
└── LICENSE
```

That's all — no `package.json`, no `lib/index.js`, no native binding. The directory exists but the package is empty. (Possibly the result of an interrupted Claude Code plugin extract, or `npm install --ignore-scripts` skipping the postinstall rebuild.)

### Why the wrapper doesn't catch it

`cli/mcp-server-wrapper.js`:

```js
const nodeModulesPath = join(PLUGIN_ROOT, 'node_modules');
if (!existsSync(nodeModulesPath)) {
  await runNpmInstall();
}
```

The check sees that `node_modules/` exists (it does — with a corrupt subset) and skips the recovery `npm install`. The server crashes immediately on import.

[...]

PR #150, opened 2026-08-12
fix(install): verify the native binding loads instead of trusting exit codes (#100, #105)

### LINKED ISSUE #100: postinstall `npm rebuild better-sqlite3` is a silent no-op on Node 25 — binding never builds, recovery hint repeats the dead command
## Summary

`scripts/postinstall.js` rebuilds the native binding with `npm rebuild better-sqlite3`. On Node 25 (macOS arm64, better-sqlite3 12.10.0) that command is a **silent no-op**: it prints `rebuilt dependencies successfully`, exits 0, and produces no `build/Release/better_sqlite3.node`. So a fresh plugin install / update ships with no loadable binding, every SessionStart sync dies with `Could not locate the bindings file`, and the MCP search/read tools return nothing.

This is the install-side root cause underneath #94. #94 is about the failure being *invisible* and explicitly assumes `npm rebuild better-sqlite3` is the working local fix — but that reporter was on Node 22 / Linux. On Node 25 the rebuild command itself doesn't work, so the postinstall can never succeed and the recovery hint it prints is dead.

## Environment

- Plugin version: 1.4.2 (also reproduced against the 1.4.2 `node_modules`)
- Plugin source: `superpowers-marketplace`
- OS: macOS (Darwin 25.5.0), arm64
- Node: v25.2.1 (NODE_MODULE_VERSION 141)
- better-sqlite3: 12.10.0
- Xcode CLT + python3 present (node-gyp can build)

## Root cause (empirical)

better-sqlite3's own install script is `prebuild-install || node-gyp rebuild --release`. The postinstall doesn't invoke that — it invokes `npm rebuild better-sqlite3`, and on this setup `npm rebuild` never runs the package's install script:

```
$ rm -f node_modules/better-sqlite3/build/Release/better_sqlite3.node

$ npm rebuild better-sqlite3 --foreground-scripts
rebuilt dependencies successfully           # exit 0

$ ls node_modules/better-sqlite3/build/Release/better_sqlite3.node
ls: ...better_sqlite3.node: No such file or directory   # nothing built
```

The two commands that DO produce a working binding, run directly in the better-sqlite3 dir:

```
$ ./node_modules/.bin/prebuild-install      # places a prebuilt, exit 0
# or
$ npm run build-release                     # = node-gyp rebuild --release, compiles locally
```

After `npm run build-release`, `new Database(':memory:')` loads and queries fine.

## Why postinstall.js can't catch it

```js
const result = spawnSync(npmBin, ['rebuild', 'better-sqlite3'], { ... });
if (result.status !== 0) { /* warn */ }
```

`npm rebuild`'s no-op exits 0, so `status !== 0` is never true — the failure is undetectable by exit code. The warning branch (which prints `Recover with: cd <plugin-dir> && npm rebuild better-sqlite3`) both never fires here AND recommends the same command that doesn't work on Node 25.

A second trap for anyone writing a health check: `require('better-sqlite3')` succeeds even with no binding, because better-sqlite3 loads the native addon lazily inside `new Database()`. So a `require()`-based preflight passes on a broken install. The binding only actually loads when a `Database` is instantiated.

## Suggested direction (not prescriptive)

The robust check is "did the binding actually load," not "did the rebuild command exit 0":

1. In postinstall, after the rebuild, **verify by instantiating** `new (require('better-sqlite3'))(':memory:')` in a child process. If it throws, fall back to `node-gyp rebuild --release` (or run the package's own `install` script / `prebuild-install`) and re-verify.
2. Prefer invoking better-sqlite3's actual install path (`prebuild-install || node-gyp rebuild --release`) over `npm rebuild`, which (at least on npm shipped with Node 25) doesn't run that package's install script.
3. Pairs naturally with #94: the same instantiate-to-verify check, run in the SessionStart hook, surfaces a real remediation instead of a silent log line.

## Reproduction

```sh
PLUGIN=~/.claude/plugins/cache/superpowers-marketplace/episodic-memory/1.4.2
rm -f "$PLUGIN/node_modules/better-sqlite3/build/Release/better_sqlite3.node"
( cd "$PLUGIN/node_modules/better-sqlite3" && npm rebuild better-sqlite3 )   # "success", builds nothing
node -e "new (require('$PLUGIN/node_modules/better-sqlite3'))(':memory:')"   # Could not locate the bindings file
( cd "$PLUGIN/node_modules/better-sqlite3" && npm run build-release )        # this actually builds it
```

## Prior art checked

- #94 — observability of the same failure; assumes `npm rebuild` is the fix (Node 22 / Linux). This issue is why that assumption doesn't hold on Node 25.
- #9 — NODE_MODULE_VERSION mismatch on newer Node.
- #19 — better-sqlite3 install pain on Node v25.

PauloACRibeiro:
Fixes #100. Addresses the probe half of #105.

Two install-path defects that each make conversation search fail silently — and that, together, keep it broken across the restarts meant to repair it.

Found on macOS arm64 / Node **v26.5.0 (NODE_MODULE_VERSION 147)** / npm 11.17.0, plugin 1.4.2 via `superpowers-marketplace`. Search had been dead for ~6 days with nothing surfaced in the UI and a clean `npm install` throughout.

---

## 1. postinstall trusted an exit code instead of the binding (#100)

`npm rebuild better-sqlite3` does not reliably run better-sqlite3's own install script, and better-sqlite3 publishes prebuilds only for the Node versions in its `engines` range (`20.x || 22.x || 23.x || 24.x || 25.x` for 12.9.0). So the rebuild can exit 0 having built nothing, or leave a binary compiled for an older ABI in place.

Here it was the ABI case — the on-disk binding was 141 while the loader required 147:

```
The module '.../better-sqlite3/build/Release/better_sqlite3.node'
was compiled against a different Node.js version using
NODE_MODULE_VERSION 141. This version of Node.js requires
NODE_MODULE_VERSION 147.
```

Both failure modes look identical from outside: a clean install, and an MCP server that dies loading the binding in a log nobody reads.

**What changed**, following the direction suggested in #100:

- **Verify by instantiating, not by requiring.** `require('better-sqlite3')` succeeds with no binding at all, because the addon loads lazily inside `new Database()`. The check now opens a database and loads `sqlite-vec` into it.
- **Verify in a child process.** A native addon cannot be unloaded, so a failed load poisons the module cache — an in-process re-check after a repair attempt would be meaningless. This is what makes the fallback below verifiable at all.
- **Actually repair.** On failure it runs better-sqlite3's real build path (`npm run build-release` = `node-gyp rebuild --release`) and re-verifies, instead of printing a recovery hint naming the command that just failed.
- **Exit code reflects the binding, not the tooling.** A noisy `npm rebuild` whose binding still loads stays non-fatal — the #95 false alarm is preserved. A clean `npm rebuild` that leaves an unloadable binding now fails the install.

Failing loudly is deliberate: a plugin whose native binding will not load has no working functionality to preserve, so a silent pass is strictly worse than a failed install.

Verified by deleting the binding outright (the #100 repro) and running postinstall — it rebuilt and re-verified at ABI 147, exit 0. And against a binding that no rebuild can fix, it attempts the source build, reports both exit codes, and exits 1:

```
  this Node        : v26.5.0 (NODE_MODULE_VERSION 147)
  binary built for : NODE_MODULE_VERSION 141  <-- ABI MISMATCH
  npm rebuild exit : 0
  source build exit: 1
```

Note `npm rebuild exit: 0` — that's the hole.

> On npm 11.17.0 `npm rebuild` *does* run the install script, so #100's "silent no-op" appears npm-version-specific. The verification is what makes that difference not matter.

---

## 2. Dependency probe reported nested packages as missing (#105)

`findMissingDeps` probed only top-level `node_modules`. npm nests `onnxruntime-node` under `@huggingface/transformers` (it pins its own range), so it read as missing on **every launch**:

```
Server stderr: Missing dependencies under node_modules: onnxruntime-node
Installing episodic-memory dependencies (first run only)...
Connection failed (CONNECT_TIMEOUT): MCP server "plugin:episodic-memory:episodic-memory"
connection timed out after 30000ms
```

`mcp-server-wrapper` therefore reran a full `npm install` (~20–22s) on every start, regularly exceeding the client's 30s connect timeout — so the server the user was waiting on never became available.

**The reinstall cannot fix what the probe detects.** npm already considers the tree complete and leaves the package nested; `npm install` ran many times here and never hoisted it. Every launch paid the cost and nothing changed.

**And the two bugs interlock:** when the connect timeout kills the wrapper mid-install it can interrupt the postinstall rebuild. That is how the broken binding from §1 survived six days of restarts that each started the repair and were each killed before finishing.

The probe now also checks one level of nesting. It deliberately stays inside the plugin's own `node_modules` rather than using `createRequire().resolve()` — the latter also walks parent directories and Node's global folders, which could report a package present that the plugin cannot actually load.

**Scope note:** this fixes the *probe*, not the resolution failure in #105. `onnxruntime-common` is not in `REQUIRED_PACKAGES`, so this change does not affect it. For the packages that are, nesting under the importer is fine — `@huggingface/transformers` resolves its own nested `onnxruntime-node`, which is confirmed working here (search returns results end to end).

Cold MCP start: **22s+ and frequently timing out → 1.6s.**

---

## Testing

- Two regression tests: a nested package must read as installed; a nested package missing its manifest must still read as missing (#95's partial-extraction case, one level down). The first fails against `main` with `expected [ 'onnxruntime-node' ] to deeply equal []`.
- Full suite: **209 passed, 35 files.**
- End to end through the real MCP transport (`initialize` → `tools/list` → `tools/call search`), not just the CLI: real results against a 6,952-exchange index.
- `sqlite-vec` confirmed loading alongside (`vec_version() = v0.1.9`).

No `src/` changes, so `dist/` is untouched and needs no rebuild.

---

## Not included

Two things I noticed but left alone, happy to split out if wanted:

1. `better-sqlite3: ^12.4.1` silently resolves to a release whose `engines` excludes the host Node, and the only signal is an `EBADENGINE` warning buried in install output. A stated supported-Node range would make this a real error rather than a surprise.
2. `CLAUDE.md` orders the release as build (step 2) → bump (step 3), so `prebuild` generates `src/version.ts` from the pre-bump `package.json` and the committed `dist/` is always one version behind. On 1.4.2 the server introduces itself as `1.4.1`. Swapping steps 2 and 3 fixes it.

### DIFF
--- a/cli/install-check.js
+++ b/cli/install-check.js
@@ -1,4 +1,4 @@
-import { existsSync } from 'fs';
+import { existsSync, readdirSync } from 'fs';
 import { join } from 'path';
 
 /**
@@ -21,8 +21,70 @@ export const REQUIRED_PACKAGES = [
   'sqlite-vec',
 ];
 
+function safeReaddir(dir) {
+  try {
+    return readdirSync(dir, { withFileTypes: true });
+  } catch {
+    return [];
+  }
+}
+
 /**
- * Return the list of required packages whose package.json is missing under
+ * True when `pkg` is installed somewhere Node's resolver will find it from the
+ * plugin root: either hoisted to the top of `node_modules` (the common case)
+ * or nested one level down under the dependent that pinned it.
+ *
+ * npm nests instead of hoisting whenever a version conflict blocks the flat
+ * layout — the same mechanism behind #105 and #135. episodic-memory hits it
+ * with onnxruntime-node, which @huggingface/transformers pins to its own range
+ * and which therefore lands at
+ * `node_modules/@huggingface/transformers/node_modules/onnxruntime-node`.
+ *
+ * A top-level-only probe calls that "missing" on every single launch, so
+ * mcp-server-wrapper reruns `npm install` every time the server starts (~20s).
+ * That routinely exceeds the client's 30s MCP connect timeout, so the server
+ * the user is waiting on never becomes available.
+ *
+ * The reinstall cannot fix what the probe detects: npm already considers the
+ * tree complete and leaves the package nested, so every launch pays the cost
+ * and nothing changes. Worse, when the connect timeout kills the wrapper
+ * mid-install it can interrupt the postinstall rebuild, which is how a broken
+ * native binding (#100) survives across restarts that were supposed to repair
+ * it.
+ *
+ * The search stays inside the plugin's own node_modules on purpose: resolving
+ * via `createRequire` would also walk parent directories and Node's global
+ * folders, which could report a package as present that the plugin cannot
+ * actually load.
+ */
+function isResolvable(nodeModules, pkg) {
+  if (existsSync(join(nodeModules, pkg, 'package.json'))) {
+    return true;
+  }
+
+  for (const entry of safeReaddir(nodeModules)) {
+    if (!entry.isDirectory()) continue;
+    const entryPath = join(nodeModules, entry.name);
+
+    // Scope directories (@foo) hold packages one level deeper.
+    const dependents = entry.name.startsWith('@')
+      ? safeReaddir(entryPath)
+          .filter(d => d.isDirectory())
+          .map(d => join(entryPath, d.name))
+      : [entryPath];
+
+    for (const dependent of dependents) {
+      if (existsSync(join(dependent, 'node_modules', pkg, 'package.json'))) {
+        return true;
+      }
+    }
+  }
+
+  return false;
+}
+
+/**
+ * Return the list of required packages that are not installed under
  * `<pluginRoot>/node_modules`. An empty array means the install looks complete;
  * a non-empty array is the diagnostic to print before re-running `npm install`.
  *
@@ -36,5 +98,5 @@ export function findMissingDeps(pluginRoot) {
   if (!existsSync(nodeModules)) {
     return REQUIRED_PACKAGES.slice();
   }
-  return REQUIRED_PACKAGES.filter(pkg => !existsSync(join(nodeModules, pkg, 'package.json')));
+  return REQUIRED_PACKAGES.filter(pkg => !isResolvable(nodeModules, pkg));
 }
--- a/scripts/postinstall.js
+++ b/scripts/postinstall.js
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 /**
- * Cross-platform postinstall: rebuild better-sqlite3's native binding against
- * the local Node version.
+ * Cross-platform postinstall: get better-sqlite3's native binding into a state
+ * where it actually loads, and refuse to report success until it does.
  *
  * Replaces the unix-only shell idiom that lived in package.json:
  *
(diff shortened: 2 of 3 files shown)


obra:
Thanks @PauloACRibeiro! Merged to main as `cbc4881` (squashed as part of a 13-PR batch, with dist rebuilt on top). This verifies the native binding actually loads at install time instead of trusting exit codes, turning a silent first-launch crash into an actionable install error.

PR #156, opened 2026-08-25
Serialize index commands with background sync

sheurich:
`index-cli` writes the same SQLite database and archive as `sync-cli`, but it does not acquire the v1.4.2 single-instance lock from #97. Manual indexing can therefore race a SessionStart sync.

This change moves the lock path into `getSyncLockPath()`, acquires it before any index command can initialize or rebuild the database, and releases it on normal exit, errors, and termination signals. Child-process tests cover contention and lock release after successful and failed commands.

Targeted tests pass: 6 tests across `test/index-cli-single-instance.test.ts` and `test/logging.test.ts`. The build, TypeScript check, and Semgrep pass. The full upstream suite remains blocked in this checkout by the existing `test/verify.test.ts` timestamp assertion and a `better-sqlite3` worker-teardown crash under Node 24.19.0.

### DIFF
--- a/dist/index-cli.js
+++ b/dist/index-cli.js
@@ -4,7 +4,30 @@ import { indexSession, indexUnprocessed, indexConversations } from './indexer.js
 import { getDbPath, getArchiveDir } from './paths.js';
 import fs from 'fs';
 import path from 'path';
+import { getSyncLockPath } from './logging.js';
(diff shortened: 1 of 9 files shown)


obra:
Thanks @sheurich! Merged to main as `f1bcef6` (squashed as part of a 13-PR batch, with dist rebuilt on top). This serializes index commands against background sync on the shared single-instance lock.

PR #157, opened 2026-08-25
Add isolated MCP integration tests

sheurich:
The suite does not boot the committed MCP server through stdio. A direct integration test can otherwise fall back to the user's configuration and conversation index.

This adds MCP tool-schema and text-search tests with temporary configuration, database, archive, projects, home, and XDG paths. The child receives no inherited `EPISODIC_MEMORY_DB_PATH`, and the search test confirms that the temporary database is used.

The 3 focused tests, build, TypeScript check, and Semgrep pass. The full upstream suite remains blocked in this checkout by the existing `test/verify.test.ts` timestamp assertion and a `better-sqlite3` worker-teardown crash under Node 24.19.0.


obra:
Thanks @sheurich! Merged to main as `8b261ea` (squashed as part of a 13-PR batch, with dist rebuilt on top). This adds genuinely isolated MCP server integration tests that spawn the built server over stdio.

PR #159, opened 2026-08-28
feat(sync): EPISODIC_MEMORY_SKIP_SUMMARIES=1 disables the summarization pass

chirayuk:
## Motivation

`episodic-memory sync` always runs the summarization pass, and there is
no way to turn it off. That pass is expensive in a way the rest of sync
is not:

- It spends the user's Claude quota. Up to 10 resume-based
  summarization calls per sync run, on every sync, forever.
- Resume-based summaries can stall on a permission prompt. Observed on
  1.4.2: of thirteen summaries in one run, two came back as text along
  the lines of "I need your approval to proceed...". For some sessions
  the resumed model continues the original conversation instead of
  summarizing it, so the sidecar ends up holding conversation text
  rather than a summary.

Search quality does not depend on any of this. `search.ts` reads the
`-summary.txt` sidecar only to decorate result output; summary text is
never embedded and never searched. So for users who want the archive
and the index but not the summarizer, the pass is pure cost.

`SyncOptions.skipSummaries` already existed and was already honored in
both the queueing and the summarizing loop — it was simply unreachable
from the CLI. This wires it to an environment variable.

## What changed

- `src/sync.ts`: new exported pure function
  `buildSyncOptionsFromEnv(env: NodeJS.ProcessEnv): SyncOptions`,
  returning `{ skipSummaries: env.EPISODIC_MEMORY_SKIP_SUMMARIES === '1' }`.
  Only the exact value `1` enables the switch, so a stray value cannot
  silently disable a feature the user still expects.
- `src/sync-cli.ts`: computes the options once and passes them to
  `syncConversations`. When the switch is on, the completion block
  prints `Summaries: skipped (EPISODIC_MEMORY_SKIP_SUMMARIES=1)` in
  place of the `Summarized:` count — a `Summarized: 0` next to a skip
  notice reads like a failure. `--background` behaviour is unchanged.
- `README.md`: the variable is documented in the API Configuration
  block, with a note that summaries are display-only.
- `dist/`: rebuilt `dist/sync.js`, `dist/sync.d.ts` and
  `dist/sync-cli.js` per CLAUDE.md's "commit `dist/` alongside `src/`".
  Two unrelated diffs that `tsc` produced were deliberately reverted so
  this PR stays scoped: `dist/version.*` (committed as 1.4.1 while
  `package.json` says 1.4.2) and the esbuild bundle
  `dist/mcp-server.js`. Both are worth a separate look.
- No `CHANGELOG.md` entry: the file has release sections only, no
  `Unreleased` convention to add to. Happy to add one under whatever
  version you want this in.

With the switch on, no `-summary.txt` is written (not even the empty
zero-exchange sentinel), the summarizer module is never imported, and
`result.summarized` stays 0. Existing sidecars are left alone, and
unsetting the variable resumes summarization where it left off.

## How tested

New `test/sync-skip-summaries.test.ts`, five tests:

- `buildSyncOptionsFromEnv` returns `skipSummaries: true` for exactly
  `'1'`.
- It returns false when the variable is unset, and for `'0'`, `'true'`,
  `'yes'`, `''`, `' 1'` and `'1 '`.
- `syncConversations` with the derived options writes no summary
  sidecar, calls no summarizer, and reports no errors. The summarizer
  is mocked the way `test/sync-error-sentinel.test.ts` mocks it.
- The negative twin: the same fixture, with options derived from an
  empty environment, does produce a sidecar and does call the
  summarizer once. Without this case the skip assertion would pass
  vacuously for a fixture that never queued a summary in the first
  place.

Full suite: 36 files, 212 tests, all passing (`vitest run`). The opt-in
`test:claude-e2e` and `test:codex-e2e` scripts were not run.

Also smoke-tested against the built `dist/sync-cli.js` with the test
path overrides pointed at a temp directory: with the variable set to
`1` the completion block prints
`Summaries: skipped (EPISODIC_MEMORY_SKIP_SUMMARIES=1)`; with it unset
it prints `Summarized: 0` as before.

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,7 +185,12 @@ export EPISODIC_MEMORY_API_TIMEOUT_MS=3000000
 
 # Override Codex binary path if needed (default: codex)
 export EPISODIC_MEMORY_CODEX_BIN=/path/to/codex
+
+# Disable the summarization pass entirely (search is unaffected)
+export EPISODIC_MEMORY_SKIP_SUMMARIES=1
 ```
+
+Only the exact value `1` disables summarization; any other value leaves it on.
 
 These settings only affect episodic-memory's summarization calls, not your interactive Claude Code or Codex sessions.
 
@@ -199,6 +204,8 @@ Codex summarization requires `codex-cli 0.130.0` or newer. If Codex app-server s
 | Embeddings | No (local Transformers.js) |
 | Search | No (local SQLite) |
 | MCP tools | No |
+
+Summaries are display-only: they decorate search results and are never embedded or searched, so `EPISODIC_MEMORY_SKIP_SUMMARIES=1` costs you that line of context and nothing else.
 
 ## Commands
 
--- a/dist/sync-cli.js
+++ b/dist/sync-cli.js
@@ -1,4 +1,4 @@
-import { syncConversations } from './sync.js';
+import { buildSyncOptionsFromEnv, syncConversations } from './sync.js';
 import { getArchiveDir, getConversationSourceDirs, getIndexDir } from './paths.js';
 import { shouldSkipReentrantSync } from './summarizer.js';
 import { initDatabase } from './db.js';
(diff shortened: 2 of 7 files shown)


obra:
Thanks @chirayuk! Merged to main as `c8d0e3f` (squashed as part of a 13-PR batch, with dist rebuilt on top). This adds EPISODIC_MEMORY_SKIP_SUMMARIES=1 to disable the summarization pass while still copying and indexing.

PR #166, opened 2026-09-09
fix(summarizer): set excludeTurns on ephemeral Codex thread/fork

### LINKED ISSUE #158: Codex summarizer: thread/fork 400s on large (paginated) threads — codex-cli 0.150 requires `excludeTurns: true`
## Summary

On current codex-cli, `thread/fork` rejects the summarizer's ephemeral fork for any **paginated** (i.e. reasonably large) thread:

```
thread/fork failed: {"code":-32600,"message":"ephemeral paginated thread/fork requires `excludeTurns: true`"}
```

`summarizeConversation()` catches this and falls back to transcript-text summarization, so nothing surfaces to the user — the only trace is `Codex summarizer unavailable, falling back to transcript text: ...` in the sync log. The effect is that the Codex-native path silently stops being used for exactly the conversations it matters most for: the long ones.

Adding `excludeTurns: true` to the fork params fixes it, and the forked thread still carries full session context (the flag governs pagination of the fork *response*, not the fork's context) — the summaries produced after the change are specific and accurate.

## Environment

- episodic-memory 1.4.2 (installed via superpowers-marketplace; identical to `main` @ `1075769`)
- codex-cli 0.150.1, ChatGPT-subscription auth, `model = "gpt-5.6-sol"` in `~/.codex/config.toml`
- macOS (darwin 25.2), node v23.7.0

## Repro

Paired A/B against the *same* Codex thread, the only difference being the flag. `summarizer.noexclude.js` is a copy of `dist/summarizer.js` with the `excludeTurns: true` line stripped:

```js
const D = '<plugin>/episodic-memory/1.4.2/dist';
const { buildCodexSummarizerCommand, runCodexCommand, buildCodexSummaryPrompt } =
  await import(D + '/' + (process.env.TEST_MOD || 'summarizer.js'));
const cmd = buildCodexSummarizerCommand({
  sessionId: process.env.TEST_SID,
  prompt: buildCodexSummaryPrompt(),
});
try { console.log('OK   ' + (await runCodexCommand(cmd)).slice(0, 200)); }
catch (e) { console.log('FAIL ' + e.message.slice(0, 200)); }
```

Results:

| thread (rollout size) | without `excludeTurns` | with `excludeTurns: true` |
|---|---|---|
| 820 KB | FAIL — `ephemeral paginated thread/fork requires \`excludeTurns: true\`` | OK — accurate summary |
| 463 KB | OK | OK |

So it is size/pagination dependent, which is presumably why it hasn't been reported: small threads still work, and the fallback hides the rest.

## Patch

`src/summarizer.ts`, in the `thread/fork` call:

```diff
         const fork = await send('thread/fork', {
           threadId: command.sessionId,
           ephemeral: true,
+          excludeTurns: true,
           sandbox: 'read-only',
           approvalPolicy: 'never',
           ...(command.model ? { model: command.model } : {}),
         });
```

Happy to open a PR if useful.

## Adjacent observations (not this bug, but they share the silent-fallback path)

From one machine's sync log, 702 Codex fork failures across 908 attempts, by class:

- **673 × `invalid_encrypted_content`** — replaying threads older than ~2 months fails because the stored reasoning items no longer verify server-side. Nothing the plugin can do about the content, but it may be worth skipping the fork (and the ~10s app-server spawn) for rollouts past some age rather than retrying them forever.
- **9 × `[ApiIdParam] Invalid 'input[0].id': 'review_rollout_user'`** — `codex review` rollouts carry synthetic message ids the API rejects on replay. These could be detected and routed straight to the transcript path.
- **15 × `The '<model>' model is not supported when using Codex with a ChatGPT account`** — residue of #98 on old threads; the fork appears to inherit the thread's recorded (now-retired) model when no override is passed, so the `undefined` default doesn't rescue those. Setting `EPISODIC_MEMORY_CODEX_MODEL` does.

All three degrade to a Claude transcript summary, so the index stays populated — the cost is quality plus a failed app-server spawn per conversation.

ada-sen:
## Summary

`thread/fork` rejects the summarizer's ephemeral fork for any **paginated** (i.e. reasonably large) Codex thread on codex-cli 0.150+:

```
thread/fork failed: {"code":-32600,"message":"ephemeral paginated thread/fork requires `excludeTurns: true`"}
```

`summarizeConversation()` catches this and silently falls back to transcript-text summarization, so the Codex-native path stops being used for exactly the conversations it matters most for: the long ones. The only trace is a `Codex summarizer unavailable, falling back to transcript text: ...` log line — nothing surfaces as an error to the user.

## Root cause

The ephemeral `thread/fork` request in `summarizeConversation()` didn't set `excludeTurns: true`. Newer codex-cli requires it for paginated (large) threads. The flag governs pagination of the fork *response*, not the fork's context — the forked thread still carries full session history with the flag set, so summaries produced after this change remain accurate.

## Fix

Add `excludeTurns: true` to the ephemeral fork params in `src/summarizer.ts`:

```diff
         const fork = await send('thread/fork', {
           threadId: command.sessionId,
           ephemeral: true,
+          excludeTurns: true,
           sandbox: 'read-only',
           approvalPolicy: 'never',
           ...(command.model ? { model: command.model } : {}),
         });
```

`dist/summarizer.js` regenerated via `npm run build` and committed alongside `src/summarizer.ts`.

## Test plan

- [x] Extended the existing `thread/fork` mock assertion in `test/summarizer-options.test.ts` to require `excludeTurns === true`, matching the existing `ephemeral`/`sandbox` checks
- [x] `npx vitest run test/summarizer-options.test.ts` — 29/29 pass
- [x] `npm test` (full suite) — 348/348 pass, 62 test files, no regressions
- [x] `npm run build` — clean

Fixes #158.

### DIFF
--- a/dist/summarizer.js
+++ b/dist/summarizer.js
@@ -560,6 +560,7 @@ export async function runCodexCommand(command) {
                 const fork = await send('thread/fork', {
                     threadId: command.sessionId,
                     ephemeral: true,
+                    excludeTurns: true,
                     sandbox: 'read-only',
                     approvalPolicy: 'never',
                     ...(command.model ? { model: command.model } : {}),
--- a/src/summarizer.ts
+++ b/src/summarizer.ts
@@ -640,6 +640,7 @@ export async function runCodexCommand(command: CodexSummarizerCommand): Promise<
         const fork = await send('thread/fork', {
           threadId: command.sessionId,
           ephemeral: true,
+          excludeTurns: true,
           sandbox: 'read-only',
           approvalPolicy: 'never',
           ...(command.model ? { model: command.model } : {}),
--- a/test/summarizer-options.test.ts
+++ b/test/summarizer-options.test.ts
@@ -143,6 +143,7 @@ describe('runCodexCommand', () => {
         if (message.method === 'thread/fork') {
           if (message.params.threadId !== 'session-123') throw new Error('wrong session id');
           if (message.params.ephemeral !== true) throw new Error('fork was not ephemeral');
+          if (message.params.excludeTurns !== true) throw new Error('fork did not set excludeTurns (required by codex-cli 0.150+ for paginated threads)');
           if (message.params.sandbox !== 'read-only') throw new Error('fork was not read-only');
           console.log(JSON.stringify({ id: message.id, result: { thread: { id: 'fork-456' } } }));
           return;


obra (approved):
Verified rather than taken on trust. Clean — merging.

**The test genuinely catches the bug.** Removed `excludeTurns: true` from `src/summarizer.ts` (i.e. reinstated the original defect) and the fork test went red on exactly the right assertion; restored, green again. That's the property that matters, and it isn't automatic — the assertion had to be added inside the fake server's `thread/fork` handler to see it.

**`dist` and `src` agree**, and a fresh `npm run build` produces no diff against the committed `dist/summarizer.js`, so the built artifact isn't carrying anything the source doesn't.

**Full suite: 62 files, 348 tests, zero failures.**

## Why this is worth landing beyond the one flag

The failure mode is the interesting part: `summarizeConversation()` catches the `-32600` and silently falls back to transcript-text summarization. So the Codex-native path quietly stopped being used for **exactly the conversations it matters most for — the long ones** — with the only trace a log line nobody reads. A degradation that announces itself as success is worse than a crash, and this repo has a few of those.

## One thing to note for later, not blocking

The fix is right and the reasoning in the description is right (`excludeTurns` governs pagination of the fork *response*, not the fork's context). But nothing in the repo pins the codex-cli version this behaviour depends on, so if a future codex-cli changes the requirement again, the same silent fallback returns and the same log line is the only signal. Worth considering whether that fallback should be loud, or at least counted, rather than swallowed. Separate change.

Also worth knowing: this repo now has CI (landed today — build, typecheck, tests across Node 22 and 24, and it runs on fork PRs). Your future PRs here will get real checks. This one predates it and shows none; I ran the equivalent locally.

PR #169, opened 2026-09-10
test(show): pin the timestamp expectation to en-US/UTC like the formatter

vicnaum:
## Problem

`formatConversationAsMarkdown` renders timestamps with `toLocaleString('en-US', { timeZone: 'UTC' })` (#120, #130), but the test in `test/show.test.ts` still builds its expectation with a bare `toLocaleString()`, i.e. the host's locale and timezone. The two only agree on en-US / UTC hosts, so the suite passes on CI and fails for anyone running it locally elsewhere:

```
Europe/Warsaw:  expected "9/19/2025, 7:34:29 PM", formatter produced "9/19/2025, 5:34:29 PM"
de_DE on UTC:   fails on the locale format
```

## Fix

Format the expectation the same way the code does. One-line test change, no runtime change.

Verified passing under `TZ=Europe/Warsaw`, `TZ=America/New_York` and `TZ=UTC`, including `LC_ALL=de_DE.UTF-8`.

https://claude.ai/code/session_01L7qt58cTm5rFNfV38EQfEN

### DIFF
--- a/test/show.test.ts
+++ b/test/show.test.ts
@@ -230,7 +230,10 @@ describe('show command - markdown formatting', () => {
     expect(markdown).toMatch(/\*\*Agent\*\*/);
     expect(markdown).toContain('Looking at your instructions');
 
-    const expectedTimestamp = new Date('2025-09-19T17:34:29.181Z').toLocaleString();
+    // Match formatConversationAsMarkdown, which pins timestamps to en-US / UTC
+    // (#120, #130). A bare toLocaleString() depends on the host locale and
+    // timezone, so the old expectation only held on en-US / UTC hosts.
+    const expectedTimestamp = new Date('2025-09-19T17:34:29.181Z').toLocaleString('en-US', { timeZone: 'UTC' });
     expect(markdown).toContain(expectedTimestamp);
   });
 


obra (approved):
Verified and landing. Thank you — this is a real portability bug and the diagnosis is exactly right.

`formatConversationAsMarkdown` pins timestamps at `src/show.ts:97`:

```js
const timestamp = new Date(msg.timestamp).toLocaleString('en-US', { timeZone: 'UTC' });
```

and the expectation used a bare `toLocaleString()`. Those agree only on an en-US/UTC host, which is what CI happens to be — so the suite was green for everyone who could run it and red for everyone who couldn't.

**Confirmed with a negative control**, because a test change that passes everywhere proves nothing on its own:

```
base 28f0933a, TZ=Europe/Warsaw  ->  1 failed | 21 passed
PR   6854e8d5, TZ=Europe/Warsaw  ->  22 passed
```

Then across the environments the fix targets — `TZ=UTC`, `TZ=America/New_York`, and `TZ=UTC LC_ALL=de_DE.UTF-8` — all 22 pass. No CI ran on the fork branch, so that local matrix is standing in for it.

**One note, not a change request.** The expectation now mirrors the implementation's exact call, so if someone changes the formatter's locale the test follows silently rather than failing. The alternative — asserting the literal `"9/19/2025, 5:34:29 PM"` — would catch that, at the cost of breaking whenever Node's ICU data shifts its formatting. For a timestamp render I think you've picked the right side of that trade; I'm flagging it only so the next person knows it was a choice.

PR #172, opened 2026-09-19
fix: move better-sqlite3 to N-API and exercise vector MCP path

Pelmog:
## Summary

- upgrade `better-sqlite3` from 12.x to `^13.0.3`, whose bundled N-API binaries work across Node module ABIs
- stop rebuilding the native addon during postinstall and fail closed when the bundled binary does not work
- strengthen postinstall verification from `vec_version()` to a real vec0 insert and nearest-neighbor query
- extend the isolated MCP integration test through initialize/list, populated text search, vector search, and archive read

## Failure reproduced

On macOS arm64 with Node 24.21.0, rebuilding `better-sqlite3` 12.11.1 from source and then running `test/integration.test.ts` deterministically aborted in `Statement::~Statement()` / `RemoveEnvironmentCleanupHook` during V8 garbage collection. This matches nodejs/node#65446 and the downstream reproduction in tobi/qmd#827.

The old 12.x binding is also ABI-specific, which is why rebuilding it under Node 26 makes it unloadable under Node 24. Version 13 moved to N-API in WiseLibs/better-sqlite3#1498 and avoids both failure modes.

## Validation

- pinned Node 24.21.0: `npm install` native vector probe passed with better-sqlite3 13.0.3
- pinned Node 24.21.0: full suite passed, 63 files / 352 tests
- Node 24.21.0: `npm run build` passed
- Node 26.8.2, using the exact same Node-24-installed `node_modules` without a rebuild: native vector probe passed
- Node 26.8.2: integration + MCP + install-policy tests passed, 21 tests

The test now fails if text or vector search returns no result, not merely if the MCP call avoids throwing.


Pelmog:
Additional real-data validation: I launched this branch under the pinned Node 24.21.0 runtime against the existing 3.4 GB episodic-memory database (without changing the installed plugin). MCP initialize/list, populated text search, embedding-backed vector search, and archive read all passed with `better-sqlite3` 13.0.3.

